PR #291, opened 2026-04-13
Up/0.8.5

patrickulrich:


src/routes/home/mod.rs
```diff
@@ -0,0 +1,1338 @@
...
+                                            {
+                                                let original_author_pk = original.pubkey.to_hex();
+                                                spawn(async move {
+                                                    let _ = crate::stores::profiles::fetch_profile(
+                                                        original_author_pk,
+                                                    )
+                                                    .await;
+                                                });
+                                            }
+                                            pending_posts.write().push(feed_item);
```
goodmorningbitcoin: This background task appends to pending_posts, but I can't find any code in this component that ever drains that buffer into feed_state. So new realtime events will increment the new-posts counter, but the visible feed never updates unless the user does a full refresh. If that is intentional, there should be an explicit merge action wired up here, otherwise this looks like a regression where realtime updates are buffered forever.

src/stores/relay/health.rs
```diff
@@ -0,0 +1,253 @@
...
+        return Vec::new();
+    }
+
+    for url in &to_remove {
+        log::warn!(
+            "Quarantining relay {} after sustained failures",
+            url
+        );
+        if let Ok(relay_url) = RelayUrl::parse(url.as_str()) {
+            let _ = client.remove_relay(relay_url).await;
```
goodmorningbitcoin, line 134: This quarantine loop permanently removes relays from the SDK pool after five 30 second polls with disconnected status, but there is no path here to re-add them later. A temporary outage or network change can therefore silently shrink the user's relay set for the rest of the session, and any user-added relay that gets quarantined is effectively gone until restart. I think this should quarantine only for routing or selection purposes, or keep a separate quarantined set with expiry/backoff and explicit re-add logic instead of calling client.remove_relay here.

src/routes/home/feed_loaders.rs
```diff
@@ -0,0 +1,671 @@
...
+    until: Option<u64>,
+) -> Result<(Vec<FeedItem>, bool), NostrBlueError> {
+    let pubkey_str = auth_store::get_pubkey().ok_or(NostrBlueError::NotAuthenticated)?;
+    log::info!(
+        "Loading following feed for {} (until: {:?})",
+        pubkey_str,
+        until
+    );
+    let contacts_future = nostr_client::fetch_contacts(pubkey_str.clone());
+    let global_future = load_global_feed(until);
```
goodmorningbitcoin: This now fetches the global feed in parallel with every following-feed request, even when the user has contacts and the following fetch succeeds. That means every normal refresh and pagination hit pays for an extra global query that is usually discarded. On slow relays this is going to add unnecessary latency and relay load. I think the global fallback should stay lazy and only execute after the contacts fetch fails or comes back empty.

src/hooks/use_relay_subscription.rs
```diff
@@ -0,0 +1,93 @@
...
+
+struct SubState {
+    sub_id: SubscriptionId,
+    client: Arc<Client>,
+}
+
+#[allow(clippy::type_complexity, clippy::arc_with_non_send_sync)]
+fn spawn_listener(client: Arc<Client>, sub_id: SubscriptionId, on_event: OnEvent) {
+    spawn(async move {
+        let mut notifications = client.notifications();
```
goodmorningbitcoin, line 17: Each call to this hook spawns its own `client.notifications()` reader loop, and every loop receives the full pool-wide notification stream forever until unmount. Once several components use this hook at the same time, every event gets fanned out through N independent receivers and each one does its own subscription-id filtering in userland. That scales poorly and is going to become expensive on busy streams. I think this wants a shared notification dispatcher or subscription manager instead of one unbounded receiver task per hook instance.

src/routes/home/mod.rs
patrickulrich: The idea is that realtime events should increment the blue banner that shows new posts since refresh with what it finds that when the users clicks that banner that the new posts are added into the feed.

goodmorningbitcoin (commented):
Second full pass complete. A few concrete issues should be addressed before merge:

1. `src/main.rs`: the stylesheet path was changed to `asset!("/public/tailwind.css")`, but this PR adds `tailwind.css` at the repo root. That breaks builds with `Asset at /public/tailwind.css doesn't exist`.

2. `src/routes/home/feed_loaders.rs`: `load_following_feed()` now kicks off a global feed fetch in parallel on every request via `futures::join!`, even when the fallback is never needed. That adds unnecessary relay load and latency for the normal path.

3. `src/hooks/use_relay_subscription.rs`: each hook instance spawns its own full `client.notifications()` loop. If multiple components mount this hook, each one consumes and filters the entire relay notification stream independently. That does not scale well and should be centralized or moved to a scoped notification handler.

I also made local fixes for these on a checked out PR branch, but this review is just about the code currently in the PR.

goodmorningbitcoin (commented):
Clarifying the actionable code issues for follow-up:

1. `src/main.rs`
   The stylesheet reference should match the asset actually added by this PR. Right now the code points at `/public/tailwind.css`, while the PR adds `tailwind.css`. The fix is to make the asset path consistent with the generated/bundled file location used by Dioxus.

2. `src/routes/home/feed_loaders.rs`
   `load_following_feed()` currently starts a global feed fetch in parallel with contact loading. That means the global path runs even when contacts load successfully and the fallback is never used. This should be changed so the global feed is fetched only on actual fallback paths, which avoids extra relay work and unnecessary latency.

3. `src/hooks/use_relay_subscription.rs`
   The current hook design starts a separate `client.notifications()` receiver loop for every mounted consumer. That duplicates work across components because each loop processes the full relay notification stream and filters client-side by subscription id. This should be refactored to a shared dispatcher or another scoped subscription mechanism so mounted components do not each consume the entire stream independently.

Suggested intent: keep subscription and fallback behavior lazy, shared where possible, and aligned with the actual asset/layout structure in the PR.

src/routes/home/mod.rs
goodmorningbitcoin: That banner behavior makes sense. I agree with the product intent.

My concern is narrower: the implementation gets there by spawning a separate `client.notifications()` receiver loop for each batch subscription. So if you have N batches, you also have N tasks each consuming and filtering the full relay notification stream independently.

I would update the recommendation to: keep the pending-post banner flow, but refactor the event delivery path so notification handling is shared/scoped rather than one full receiver loop per batch. The buffering UX is fine, the notification fan-out architecture is the part that looks expensive and fragile.

src/components/blobbi/social/breeding_modal.rs
```diff
@@ -0,0 +1,146 @@
...
+                            } else {
+                                "w-full py-2 bg-primary text-primary-foreground rounded-lg text-sm hover:opacity-80 transition"
+                            },
+                            disabled: breeding() || partner_id().is_empty() || partner_owner().is_empty(),
+                            onclick: move |_| {
+                                let blobbi_d = blobbi.d.clone();
+                                let blobbi_name = blobbi.display_name().to_string();
+                                let pid = partner_id();
+                                let powner = partner_owner();
+                                spawn(async move {
```
goodmorningbitcoin, line 111: This flow still treats publish as fire-and-forget. `send_event_builder(...)` errors are discarded, but the UI immediately shows success / closes anyway. That can falsely tell the user a breeding or photo/share action was published when nothing actually reached a relay. At minimum, surface the error and avoid committing success UI until the send result is known.

src/hooks/use_relay_subscription.rs
```diff
@@ -0,0 +1,93 @@
+use crate::stores::nostr_client;
+use dioxus::prelude::*;
+use nostr_sdk::prelude::*;
+use nostr_sdk::{Client, RelayPoolNotification, SubscribeAutoCloseOptions, SubscriptionId};
+use std::sync::{Arc, Mutex};
+
+type OnEvent = Arc<Mutex<Box<dyn FnMut(&nostr::Event)>>>;
+
+struct SubState {
+    sub_id: SubscriptionId,
```
goodmorningbitcoin, line 10: This hook still starts a fresh client.notifications() receiver for every subscription instance. That means each mounted consumer is pulling the full relay notification stream and filtering locally by SubscriptionId, which scales badly and risks duplicated work or stream contention as more components subscribe. I still think this needs a shared dispatcher or the client's filtered notification handling instead of one raw receiver per hook.

goodmorningbitcoin (commented):
Did another pass on the updated head (2591e8512e09b8f2c1366fc5b8d5094b56deaddc) and left inline comments for the remaining issues I still think should be addressed before merge. Highest confidence items from this pass: (1) the stylesheet asset path in src/main.rs is still breaking cargo check on the current head, (2) use_relay_subscription still creates one raw notifications() consumer per subscription instance, and (3) Blobbi publish flows still report success even if send_event_builder fails.

PR #292, opened 2026-04-15
Release 0.8.5: Android refinements, feed architecture improvements, AI tools, Cashu feature-gating, and bug fixes

patrickulrich:
- **Android integrations & theming** — Native routing, theming, and radio stream buffering with fallback logic
- **Home feed refactor** — Modularized feed into separate modules, scroll restoration, viewport engagement tracking, adaptive since-window for following filter
- **Note page overhaul** — Batched engagement metrics, iterative ancestor fetching with relay hints, inbox relay targeting
- **AI chat Nostr tools** — 15 native tools for reading profiles, notes, events, interactions, and search
- **Cashu wallet feature-gate** — NIP-60/61 wallet behind `cashu` feature flag for Play Store compliance
- **External signer optimization** — Reduced NIP-07/NIP-55/NIP-46 calls with caching and deduplication
- **Relay subscription hook** — Migrated all relay subscriptions to unified `use_relay_subscription` hook
- **Starter pack publishing** — Fixed publishing, caching, and added calendar event dedup with real-time updates
- **Reply publishing** — Uses SDK `text_note_reply`, broadcast fallback, error toasts, added Blobbi feature

## Bug Fixes
- Fix hashtag feed not reacting to tag navigation via sidebar (use_reactive! pattern)
- Fix note detail page not loading replies when navigating between hot posts (state clearing + generation counter)
- Fix notification fan-out and eager global feed fetch in home feed
- Fix reply publishing with proper SDK method and broadcast fallback
- Fix Android radio stream buffering with fallback logic

## Commits (13)
- 88dbfce7 Refine Android integrations, routing, and native theming
- a256fbc7 Migrate all relay subscriptions to use_relay_subscription hook
- 24830636 Refactor home feed into modules, fix scroll restoration, add viewport engagement and adaptive since-window
- 560959f0 Overhaul Note page loading: batch engagement, iterative ancestors, relay hints, inbox relay targeting
- de5460ec Feature-gate Cashu wallet (NIP-60/NIP-61) behind 'cashu' feature flag for Play Store compliance
- 66e9331a Reduce external signer calls (NIP-07/NIP-55/NIP-46) with caching and deduplication
- 56d346da Add 15 native Nostr tools to AI chat for reading profiles, notes, events, interactions, search, and more
- 76f61396 Fix Android radio stream buffering with fallback logic, show share modal toolbar on mobile, bump to 0.8.5
- 10714727 Fix reply publishing: use SDK text_note_reply, add broadcast fallback, toast errors, add Blobbi feature
- b5d34ddc Fix starter pack publishing and caching, add calendar event dedup and real-time updates
- eac1cd18 Fix notification fan-out and eager global feed fetch in home feed
- 2591e851 Fix reactivity bugs in hashtag feed and note detail page
- 8a865dee Merge pull request #291 from patrickulrich/up/0.8.5


src/hooks/use_relay_subscription.rs
```diff
@@ -0,0 +1,93 @@
+use crate::stores::nostr_client;
+use dioxus::prelude::*;
+use nostr_sdk::prelude::*;
+use nostr_sdk::{Client, RelayPoolNotification, SubscribeAutoCloseOptions, SubscriptionId};
+use std::sync::{Arc, Mutex};
+
+type OnEvent = Arc<Mutex<Box<dyn FnMut(&nostr::Event)>>>;
+
+struct SubState {
+    sub_id: SubscriptionId,
```
goodmorningbitcoin, line 11: This hook still starts a fresh `client.notifications()` receiver for every subscription instance. That means each mounted consumer is pulling the full relay notification stream and filtering locally by `SubscriptionId`, which scales badly and risks duplicated work or stream contention as more components subscribe. I still think this needs a shared dispatcher or the client's filtered notification handling instead of one raw receiver per hook.

src/components/blobbi/shop/inventory_modal.rs
```diff
@@ -0,0 +1,168 @@
...
+    match stage {
+        crate::utils::nip_bb::BlobbiStage::Egg => false,
+        _ => !matches!(item_id, "ball" | "teddy" | "kite" | "game" if false),
+    }
+}
+
+async fn use_item_on_blobbi(
+    blobbi: &mut BlobbiCompanion,
+    item_id: &str,
+) -> Result<(), String> {
```
goodmorningbitcoin, line 131: There is still a state-consistency problem here: the inventory count is decremented in the local `profile` before either publish completes, but `blobbi_profile_store::set_profile(profile)` only runs after both network writes succeed. If `publish_blobbi_state(...)` succeeds and `publish_profile(...)` fails, the pet state is already persisted while the inventory change is dropped locally and remotely. That lets a single item be consumed without the corresponding storage decrement ever being committed. This needs an atomic/rollback strategy, or at least the profile publish should happen before the state change is treated as successful.

src/components/blobbi/social/breeding_modal.rs
```diff
@@ -0,0 +1,146 @@
...
+                                        &pid,
+                                        "self",
+                                        &powner,
+                                        success,
+                                        Some(&offspring_id),
+                                        content,
+                                    );
+
+                                    if let Some(client) = nostr_client::NOSTR_CLIENT.read().as_ref() {
+                                        let _ = client.send_event_builder(event).await;
```
goodmorningbitcoin: This flow still treats publish as fire-and-forget. `send_event_builder(...)` errors are discarded, but the UI immediately shows success anyway. That can falsely tell the user a breeding action was published when nothing actually reached a relay. At minimum, surface the error and avoid committing success UI until the send result is known.

goodmorningbitcoin (commented):
Did a focused review pass on PR 292 and left inline comments for the issues I still think should be addressed before merge. Highest confidence findings from this pass: (1) `src/main.rs` still references `asset!("/public/tailwind.css")`, and local `cargo check` on head `8a865dee0babbf710961c119fa852a1f3049868f` is still failing on that exact line, (2) `use_relay_subscription` still creates one raw `client.notifications()` consumer per subscription instance, and (3) the new Blobbi flows still have publish/state consistency problems, including optimistic success on failed sends and item consumption that can persist pet state without persisting the corresponding inventory decrement.

goodmorningbitcoin (commented):
Did the requested second pass, focused on AI chat and Cashu.

I found two additional issues that I was not able to anchor inline because GitHub rejected the exact line mapping on the current diff, so I am leaving them here in the review summary instead:

1. In `src/routes/ai_chat.rs`, `apply_chat_response()` feeds tool results back into the follow-up request as synthetic `ChatRole::User` messages (`[Tool "..." returned: ...]`). That changes the conversation semantics, because the model sees tool output as fresh user instructions instead of tool/assistant context. This can skew the next response and makes imperative tool output look like user intent.

2. In `src/stores/cashu/events.rs`, incremental token sync advances `last_token_sync` to `Timestamp::now()` after a fetch pass. Because relay visibility is eventual and the next query uses `since(last_sync_ts)`, this can skip older-but-late-arriving token events permanently. The watermark should usually be derived from observed event timestamps with overlap, or use a deliberate lookback window, instead of jumping straight to local wall clock time.

src/stores/notification_dispatcher.rs
```diff
@@ -0,0 +1,147 @@
...
+            DispatcherHandle {
+                sub_id,
+                callback_id,
+                client: dispatcher.client.clone(),
+            },
+            rx,
+        ))
+    }
+
+    pub async fn unregister(self) {
```
goodmorningbitcoin, line 141: This dispatcher fixes the fan-out problem, but `unregister()` still tears down the underlying relay subscription unconditionally via `client.unsubscribe(&self.sub_id)`. If multiple components are sharing the same `SubscriptionId` through the dispatcher, the first one to unmount will unsubscribe the relay stream for all of them. You need reference-counted ownership here, or only unsubscribe when the last callback for that `sub_id` is removed.

goodmorningbitcoin (commented):
Re-audited the newly pushed head `0b49635b5f4e5562d3d2ce7ef05f8e46bbf4baca`.

Good news:
- AI chat tool-call follow-up semantics look corrected.
- Cashu incremental sync watermark now has a lookback window.
- Inventory flow now publishes profile before Blobbi state and attempts rollback if the second publish fails.

Remaining issues from this pass:
1. `src/main.rs` still references `asset!("/public/tailwind.css")` on the current head, so the build break still appears to be present.
2. The new notification dispatcher still unsubscribes the underlying relay subscription on every callback removal, which can break other listeners sharing the same `SubscriptionId`.
3. `src/components/blobbi/social/photo_modal.rs` still fire-and-forgets `send_event_builder(...)` and closes immediately, so the user can still be told a share succeeded when nothing was published.

PR #293, opened 2026-04-17
Up/0.8.6

patrickulrich:


goodmorningbitcoin:
Audit pass on PR 293. Two concrete issues.

1. Build is currently broken on the checked out PR head `f98499a91ceb05446fa1b52cf77834dbe5c733c3`.

```text
error: Asset at /public/tailwind.css doesn't exist
   --> src/main.rs:130:38
130 |         document::Stylesheet { href: asset!("/public/tailwind.css") }
```

That makes this a hard stop until the asset path or bundling config is fixed.

2. `PersistentMusicPlayer` can leave the old audio element playing when the track is cleared.

In `src/components/music/player/mod.rs`, the reactive effect increments `native_bind_token`, then in the `current_track == None` branch it uses `previous_bind_token` for the cleanup script. That means the cleanup path only runs if the DOM still holds the old token. If a prior bind already updated `audio.dataset.bindToken` to the new token, the guard fails and the player skips `pause() / src = "" / detach()`. Result: closing or clearing the player can leave background audio running.

The cleanup branch should use the current token consistently, or not gate teardown on the stale token at all.

src/components/music/player/mod.rs
```diff
@@ -0,0 +1,621 @@
...
+        move |(current_track, is_playing)| {
+            native_source_bound.set(false);
+            let previous_bind_token = *native_bind_token.peek();
+            let bind_token = native_bind_token.with_mut(|token| {
+                *token = token.wrapping_add(1);
+                *token
+            });
+
+            let Some(track) = current_track.as_ref() else {
+                let bind_token = previous_bind_token;
```
goodmorningbitcoin: The teardown path here is gated on `previous_bind_token`, but the effect has already incremented `native_bind_token` above. If a prior bind updated `audio.dataset.bindToken` to the new token, this branch returns before `pause()`, clearing `src`, or detaching HLS. In practice that can leave audio playing after the player is closed or the current track is cleared. Use the current token consistently here, or skip the token guard for teardown.

goodmorningbitcoin (commented):
Found one inline correctness issue and one non-inline blocker.

- Inline comment left on `src/components/music/player/mod.rs` for the stale bind-token teardown path that can leave audio playing after close/clear.
- Also worth fixing before merge: the PR head currently fails to build in my checkout with:

```text
error: Asset at /public/tailwind.css doesn't exist
   --> src/main.rs:130:38
130 |         document::Stylesheet { href: asset!("/public/tailwind.css") }
```

I could not leave that one as an inline review comment because `src/main.rs` is not part of this PR diff, but it is still a real blocker on the checked out head `f98499a91ceb05446fa1b52cf77834dbe5c733c3`.

src/components/music/player/player_expanded.rs
```diff
@@ -0,0 +1,357 @@
...
+                // Seek bar (draggable)
+                if !track.is_live_stream {
+                    div { class: "w-full mb-4",
+                        div {
+                            class: "relative h-6 flex items-center cursor-pointer touch-none",
+                            onpointerdown: move |evt: Event<PointerData>| {
+                                is_scrubbing.set(true);
+                                let client_x = evt.client_coordinates().x;
+                                let el = evt.data.element_coordinates();
+                                let width = el.x.max(1.0);
```
goodmorningbitcoin: This scrubber math is using element_coordinates().x as if it were the element width. It is the pointer's local X coordinate, not the bar width, so client_x / width produces unstable percentages and usually jumps straight to ~100% once the pointer moves away from the left edge. The bar version above correctly uses the element's bounding rect. This path should do the same or expanded-mode seeking will be badly inaccurate.

src/components/music/player/player_floating.rs
```diff
@@ -0,0 +1,161 @@
...
+                is_dragging.set(false);
+
+                let distance = *total_drag_distance.read();
+                if distance < 5.0 {
+                    music_player::restore_from_floating();
+                    return;
+                }
+
+                let (cx, cy) = pos();
+                let viewport_height = 800.0_f64;
```
goodmorningbitcoin, line 88: This uses a hard-coded 800px viewport height to decide whether the pill should dismiss. On small screens that makes close-by-drag nearly impossible, and on tall screens it can dismiss far too early. The threshold should come from the actual viewport height (or container bounds), otherwise the floating player behavior changes unpredictably across devices.

goodmorningbitcoin (commented):
Fresh pass on the new head `bff6db89647e43339a469ccfcefe038640d6cd82`.

I re-checked the previously reported issues first:
- the stale bind-token teardown bug in `PersistentMusicPlayer` is fixed
- the Tailwind asset build break is fixed by switching to `option_asset!`

New concrete issues left inline on this pass:
- expanded player scrubber is using `element_coordinates().x` as if it were element width, which makes seeking inaccurate / jumpy
- floating player dismiss logic uses a hard-coded 800px viewport height, so drag-to-dismiss behavior varies incorrectly across devices

`cargo check` is still running in my local review checkout. If it turns up anything else real, I’ll add it separately.

src/components/music/player/player_expanded.rs
```diff
@@ -0,0 +1,379 @@
...
+                }
+
+                // Seek bar (draggable)
+                if !track.is_live_stream {
+                    div { class: "w-full mb-4",
+                        div {
+                            id: "expanded-seek-bar",
+                            class: "relative h-6 flex items-center cursor-pointer touch-none",
+                            onpointerdown: move |evt: Event<PointerData>| {
+                                is_scrubbing.set(true);
```
goodmorningbitcoin: This still has a race on drag start.  marks scrubbing active immediately, but the actual bar bounds are loaded asynchronously via . Any  events that arrive before that future completes will use the default  and , which makes the scrubber jump to 0% or 100% on fast drags. The bounds need to be available synchronously before enabling scrubbing, or the move handler needs to ignore input until they are initialized.

goodmorningbitcoin (commented):
I dug back into the latest head `5317612515e2c5019fc276c2d35b117644296b4d`.

Only code delta from the last reviewed head was `player_expanded.rs`, so I focused there first. I left one new inline comment for a remaining drag-start race in the expanded scrubber implementation: scrubbing is enabled before the async bounds lookup completes, so fast pointer moves can still compute against the default `left=0,width=1` values and jump the thumb to the extremes.

I also have `cargo check` running on this head in my local checkout. If it turns up another concrete regression, I’ll add it separately.

src/components/music/player/player_expanded.rs
```diff
@@ -0,0 +1,379 @@
...
+
+                // Seek bar (draggable)
+                if !track.is_live_stream {
+                    div { class: "w-full mb-4",
+                        div {
+                            id: "expanded-seek-bar",
+                            class: "relative h-6 flex items-center cursor-pointer touch-none",
+                            onpointerdown: move |evt: Event<PointerData>| {
+                                let client_x = evt.client_coordinates().x;
+                                spawn(async move {
```
goodmorningbitcoin, line 177: This avoids the fast-drag race, but there is still a tap/release race: if the user releases before the async  finishes,  runs first with no scrub position, then the future comes back and sets both  and  afterward. That can leave the component stuck in scrubbing mode and the next pointer move will jump the thumb unexpectedly. You need a cancellation/gesture token, or to ignore async results once the gesture has already ended.

PR #294, opened 2026-04-18
Release v0.8.6: Play Store build, Bible commentary & offline, edit proposals, expanded music player, ExoPlayer migration, and sidebar discovery

patrickulrich:
This release adds Google Play Store support (without Cashu), a fully reworked Bible experience with offline translations and classic commentaries, NIP-41 edit proposals, a three-mode music player (bar/expanded/floating), Android ExoPlayer migration, and enhanced sidebar discovery with Nostrarchives integration.
---
## Play Store Build Target
- New `playstore` Cargo feature alias that enables `mobile_platform` (NIP-55 Android signer, native audio, mobile downloads) **without** the `cashu` feature, making the build compliant with Google Play policies
- `scripts/google-play.sh` produces an AAB with `ANDROID_FEATURES=playstore`
- Proguard rules, network security config for cleartext HTTP media streams, and ExoPlayer Gradle dependencies (`media3-exoplayer:1.9.0`, HLS, session modules) in `Dioxus.toml`
## Android ExoPlayer/Media3 Migration
- Replaced raw `MediaPlayer` with ExoPlayer in `NativeAudioBridge.kt` — all player operations are marshaled to the main thread via `CountDownLatch`
- `MediaPlaybackService` as a foreground service with `mediaPlayback` type, `MediaSession` for lock screen/notification controls
- Typed `NativeQueueItem` data class with full metadata (title, artist, album, art URL, duration, live/podcast flags)
- Supports standard media actions: PLAY, PAUSE, NEXT, PREVIOUS, STOP via `attachService`/`detachService` lifecycle
## Expanded/Floating Music Player
- Split monolithic `player.rs` into three submodules: `player_bar`, `player_expanded`, `player_floating`
- `PersistentMusicPlayer` orchestrator renders the correct view based on `PlayerViewMode` (Bar / Expanded / Floating)
- **Floating player**: draggable pill with pointer events, drag-to-dismiss (bottom threshold with animated red target zone), CSS transitions for smooth repositioning, respects safe area insets
- **Expanded player**: full-screen view with album art, track info, seek bar with pointer-event-driven scrubbing using `getBoundingClientRect` for accurate percentages
- Android hardware back button support via `close_topmost_mobile_overlay` — collapses expanded player to bar mode before navigating away
### Scrubber Bug Fixes (from PR review)
- Fixed percentage math: replaced `element_coordinates().x` (pointer's local offset, NOT element width) with `getBoundingClientRect()` — uses the same `(clientX - rect.left) / rect.width` formula as the bar version
- Fixed race condition: `is_scrubbing` is now set **inside** the async callback after bounds are cached, and `pointermove` is gated on `seek_bar_width > 1.0` to ignore input until bounds are initialized
- Fixed stale bind token in cleanup path: when `current_track == None`, audio is now unconditionally paused/cleared instead of gating on a token that could be stale from a prior bind
## Bible Commentary & Offline Support
- **Commentary panel**: dropdown selecting from six classic commentaries (Matthew Henry, Adam Clarke, Jamieson-Fausset-Brown, John Gill, Keil & Delitzsch OT-only, Tyndale Study Notes), fetched from HelloAO Bible API with optional introduction sections
- **Cross-reference panel**: verse-grouped cross-references with navigable links covering all 66 books
- **Offline translations**: `BibleOfflineStorage` trait with `IndexedDbBibleStorage` (web) and `SqliteBibleStorage` (native) backends — stores complete translations as JSON blobs, versioned schema with auto-migration from older split-store formats
- **Translation picker modal**: full-screen overlay organizing translations into Downloaded (favorites first), Favorites, Recommended (quick-select pills), and All Languages (grouped by language, collapsible) with search across ID/name/language fields
- **Download progress button**: compact stateful indicator with three states — downloading (spinner), downloaded (green checkmark with remove confirmation), and not-downloaded (cloud-download icon)
- Expanded chapter view with AI context generation for verse-by-verse commentary queries
## Edit Proposals (NIP-41 / Kind-1010)
- `EditPostView` shared editor supporting two modes: **direct editing** (own posts) and **edit proposals** (Kind 1010 for suggesting edits to others' posts), toggled via `is_proposal` prop
- `EditProposalCard` renders incoming Kind 1010 proposals on the original post with proposer info, summary tag, and "Accept Suggestion" button (visible only to the original author) that pre-fills the editor with proposal content
- `EditStatus` badge shows "Edited X ago" with a dialog displaying full edit history chronologically (Original, Edit 1, Edit 2, etc.)
- `edit_cache` global store deduplicates by event ID, sorts by timestamp, and bumps `EDIT_VERSION` signal for reactive UI refresh
- Profile feed support: processes kind-1010 events in aggregation counting and streaming pipelines, updates profile stats
## Nostrarchives Sidebar Discovery
- Right sidebar Hot Posts and Trends panels now cycle through three sources: NostrWine → Ditto → **Nostrarchives** (new)
- Nostrarchives API integration (`api.nostrarchives.com`): `/v1/notes/trending` for hot posts, `/v1/hashtags/trending` for trending hashtags
- Hot posts are filtered against the user's mute list before rendering
- Source selection persisted to local storage; Ditto trends include SVG sparkline mini-charts showing 7-day activity
## AI Provider Enhancements
- Three provider kinds: **PPQ** (built-in, managed key), **OpenAI-compatible** (Bearer token), **Anthropic** (X-Api-Key header auth)
- Provider management with validated HTTPS-only base URLs (with private IP exception), credential stripping, per-provider model selection
- Queued save system: batches rapid changes, coalesces to latest snapshot, serializes to IndexedDB
- Legacy migration path for old "shakespeare" provider ID → "ppq"
- Multi-turn tool use support, streaming, multi-part messages including image URLs
## Other Additions
- **`TextWithLinks`**: lightweight inline renderer that linkifies HTTP URLs and Nostr URIs (`nostr:npub1...`, `nostr:note1...`, etc.) with trailing punctuation cleanup — simpler alternative to `RichContent` for basic contexts
- **`option_asset!` for tailwind CSS**: eliminates compile-time failure on clean checkouts by returning `None` when the generated CSS file doesn't exist, while preserving the full asset pipeline (cache busting, optimization) when it does


src/components/music/player/player_expanded.rs
```diff
@@ -0,0 +1,379 @@
...
+                            "Buffering..."
+                        }
+                    }
+                }
+
+                // Seek bar (draggable)
+                if !track.is_live_stream {
+                    div { class: "w-full mb-4",
+                        div {
+                            id: "expanded-seek-bar",
```
goodmorningbitcoin, line 174: This still has the tap/release race from the previous PR: if the user releases before the async bounds lookup completes, `onpointerup` clears the gesture first, then the async callback comes back and sets `scrub_position` and `is_scrubbing` afterward. That leaves stale scrubbing state behind and the next pointer move can jump the thumb unexpectedly. This needs a gesture token / cancellation check before applying the async result.

src/components/music/player/player_floating.rs
```diff
@@ -0,0 +1,161 @@
...
+                is_dragging.set(false);
+
+                let distance = *total_drag_distance.read();
+                if distance < 5.0 {
+                    music_player::restore_from_floating();
+                    return;
+                }
+
+                let (cx, cy) = pos();
+                let viewport_height = 800.0_f64;
```
goodmorningbitcoin: This dismissal logic is still hard-coded to an 800px viewport. On shorter screens that makes drag-to-close nearly unreachable, and on taller screens it dismisses too early. The threshold should be derived from the actual viewport or container height, otherwise the floating player behaves inconsistently across devices.

src/components/music/player/player_expanded.rs
```diff
@@ -0,0 +1,387 @@
...
+                            class: "relative h-6 flex items-center cursor-pointer touch-none",
+                            onpointerdown: move |evt: Event<PointerData>| {
```
goodmorningbitcoin, line 176: This gesture cancellation fix still misses the `pointercancel` path. The async callback now bails on `pointerup` and `pointerleave`, but there is no `onpointercancel` handler here to bump `gesture_id` and clear scrubbing state. On touch devices, cancelled gestures can still resurrect stale async results and leave the scrubber in an inconsistent state.

src/components/music/player/player_floating.rs
```diff
@@ -0,0 +1,170 @@
...
+
+                #[cfg(feature = "web")]
+                let dismiss_threshold = {
+                    let vh = web_sys::window()
+                        .and_then(|w| w.inner_height().ok())
+                        .and_then(|h| h.as_f64())
+                        .unwrap_or(800.0);
+                    vh - 120.0
+                };
+                #[cfg(not(feature = "web"))]
```
goodmorningbitcoin, line 97: The web path is fixed, but native still falls back to the same hard-coded `800 - 120` threshold. This PR adds Android/ExoPlayer work, so that fallback matters. On native devices with shorter or taller screens, drag-to-dismiss will still feel wrong. The threshold should come from the actual viewport/container height on native too, or at least from a platform query instead of a fixed constant.

PR #295, opened 2026-04-22
v0.8.7: Add publish queue, NIP-36 sensitive content, Verts feed, and wiki routing

patrickulrich:
- **Publish queue** — New offline-first publish queue with persistence, retry logic, and relay-level tracking. All event publishing (notes, replies, articles, zaps, polls, etc.) now routes through the queue, providing reliable delivery with per-relay success/failure feedback and a publish queue indicator in the UI.
- **NIP-36 sensitive content** — Full content-warning tag support across all event renderers (notes, photos, videos, articles, polls, DMs, highlights, bookmarks, etc.) with a `filter: blur()` overlay and click-to-reveal. Composer gets a "Mark as sensitive" toggle with optional reason field. Settings toggle to auto-reveal.
- **Verts feed** — Fix Verts feed switching and add Verts sidebar item for video discovery.
- **Wiki routing** — Add slug-based wiki page routing (`/wiki/:slug`) and improve wiki navigation.
- **Composer refactor** — Extract shared `UseComposerEditor` hook and `ComposerBody` component from NoteComposer/ReplyComposer. Add draft auto-save, discard modal, and mention autocomplete to all composers.
- **NIP-65 relay connections** — Connect all NIP-65 read/write relays at login for reliable message delivery to DM and note relays.
- **Version bump** to 0.8.7


src/hooks/use_composer_editor.rs
```diff
@@ -0,0 +1,188 @@
...
+        } else if *show_warning.read() {
+            "text-yellow-500"
+        } else {
+            "text-gray-500"
+        }
+    });
+
+    let draft_ctx = config.draft_context.clone();
+    let draft_context_signal: Signal<Option<String>> = use_signal(move || draft_ctx);
+    use_effect(move || {
```
goodmorningbitcoin, line 58: **Draft auto-save fires on every keystroke (no debounce)**

The use_effect hook reads content on every render, which triggers save_note_draft to storage on every character typed. This causes excessive writes during active typing.

Consider adding a debounce before persisting, or saving on blur/visibility change instead.

src/stores/publish_queue/persistence.rs
```diff
@@ -0,0 +1,79 @@
...
+    }
+    Ok(())
+}
+
+#[cfg(not(all(target_arch = "wasm32", feature = "web", not(feature = "native"))))]
+pub async fn add_queued_event(_event: &QueuedEvent) -> Result<(), String> {
+    Ok(())
+}
+
+#[cfg(not(all(target_arch = "wasm32", feature = "web", not(feature = "native"))))]
```
goodmorningbitcoin: **Non-web platforms have no-op persistence — queued events lost on restart**

All persistence functions for non-wasm targets are stub no-ops that silently return Ok(()). On native/desktop/mobile builds, the publish queue exists only in memory. If the app crashes or is closed while events are pending, they are permanently lost.

At minimum, consider logging a warning when persistence is a no-op, or using file-backed storage as a fallback for native targets.

src/stores/publish_queue/types.rs
```diff
@@ -0,0 +1,74 @@
...
+
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct QueuedEvent {
+    pub id: String,
+    pub event_json: String,
+    pub event_type: QueueEventType,
+    pub event_id: String,
+    pub pubkey: String,
+    pub status: QueueEventStatus,
+    pub target_relays: Option<Vec<String>>,
```
goodmorningbitcoin, line 51: **Kind() and d_tag() re-deserialize event JSON on every call**

These methods call serde_json::from_str on self.event_json each time they are invoked. The coalescing logic in enqueue() calls both kind() and d_tag() for every existing event in the queue during the linear scan. For a queue with N events, this means 2N JSON deserializations per enqueue call.

Consider caching the parsed Kind and d_tag in QueuedEvent, or parsing once when the event is first enqueued.

src/components/note_composer.rs
```diff
@@ -1,243 +1,279 @@
...
+                spawn(async move {
+                    match publish_note_tracked(content_value, Vec::new(), content_warning.clone()).await {
+                        Ok(_) => {
+                            show_queued_toast(toast_api, "Note");
+                            editor.clear();
+                            editor.clear_draft();
+                            nav.push(crate::routes::Route::Home {
+                                list: String::new(),
+                            });
+                        }
```
goodmorningbitcoin, line 112: **FullPage publish error path has no user feedback**

In the NoteMode::FullPage branch, if publish_note_tracked returns Err(e), only a log::error is emitted. The user sees no toast, no error message — just the spinner stopping via is_publishing.set(false) outside the match. Compare with the Inline branch which shows feedback via publish_feedback.

The user needs to know the publish failed. Consider adding a toast_api.error() call in the Err branch, similar to the Inline error handling.

src/stores/cashu/init.rs
```diff
@@ -384,28 +379,20 @@ pub async fn create_wallet(mints: Vec<String>) -> Result<(), String> {
         .map_err(|e| format!("Failed to encrypt wallet data: {}", e))?;
```
goodmorningbitcoin, line 382: **Wallet create_wallet() sets Ready state before confirming relay acceptance**

The old code only set WALLET_STATE and WALLET_STATUS to Ready after verifying that at least one relay accepted the wallet event. The new code sets these immediately after enqueue(), before the event has actually been published to any relay.

If the publish queue processing fails (all relays reject, network error, max retries exceeded), the user will see a Ready wallet that was never actually persisted to the network. The wallet state would be inconsistent: the app thinks the wallet exists, but no relay has the event.

Consider using enqueue_and_await() here instead of enqueue() so the Ready state is only set after successful relay acceptance.

src/utils/nips/nip36.rs
```diff
@@ -0,0 +1,11 @@
+use nostr_sdk::{TagKind, TagStandard};
+
+pub fn get_content_warning(tags: &nostr_sdk::Tags) -> Option<Option<String>> {
```
goodmorningbitcoin, line 3: **get_content_warning() returns Option<Option<String>> — confusing double-Option**

The outer Option indicates whether a content-warning tag exists; the inner Option indicates whether a reason string was provided. This means callers must use if let Some(reason) = content_warning { if let Some(reason_text) = reason { ... } } pattern.

Looking at the callers (photo_card.rs, video_card.rs), they do:
  if let Some(reason) = content_warning.clone() {
      SensitiveContent { reason, ... }
  }

But SensitiveContent expects reason: Option<String>, and here reason is actually Option<String> (the inner Option from the double-Option). This means:
  - If tag exists WITH reason: Some(reason) -> shows the reason text (correct)
  - If tag exists WITHOUT reason: None -> SensitiveContent wraps content but shows default "Sensitive Content" text (correct behavior but unclear intent)
  - If no tag exists: the else branch renders without SensitiveContent (correct)

The API works, but the double-Option is easy to misuse. Consider returning Option<String> directly where None means "no content warning" and Some(String::new()) means "warning with no reason".

src/services/git_hosting/discussions.rs
```diff
@@ -98,11 +103,10 @@ pub async fn publish_discussion_comment(
...
         builder = builder.tag(Tag::coordinate(coord.clone(), None));
     }
-    let output = client
-        .send_event_builder(crate::utils::nips::nip89::tag_event_builder(builder))
+    let event = crate::stores::publish_queue::signing::sign_event_builder(builder)
         .await
-        .map_err(|e| format!("Failed to publish comment: {}", e))?;
-    Ok(*output.id())
+        .map_err(|e| format!("Failed to sign: {}", e))?;
+    Ok(event.id)
```
goodmorningbitcoin: **Git hosting discussion_comment and issue_comment skip the publish queue**

publish_discussion_comment and publish_issue_comment sign the event but then immediately return the event ID without enqueueing it. Unlike all the other git hosting functions (issues, PRs, releases, repos, bounties) which now use publish_queue::enqueue(), these two comment functions bypass the queue entirely.

This means comment events are fire-and-forget from the signing step — they are signed but never sent to any relay through the queue. If there was a relay send step that got removed, these comments will silently never appear on the network.

src/services/git_hosting/issues.rs
```diff
@@ -215,11 +225,10 @@ pub async fn publish_issue_comment(
...
         builder = builder.tag(Tag::coordinate(coord.clone(), None));
     }
-    let output = client
-        .send_event_builder(crate::utils::nips::nip89::tag_event_builder(builder))
+    let output = crate::stores::publish_queue::signing::sign_event_builder(builder)
         .await
-        .map_err(|e| format!("Failed to publish comment: {}", e))?;
-    Ok(*output.id())
+        .map_err(|e| format!("Failed to sign: {}", e))?;
+    Ok(output.id)
```
goodmorningbitcoin: **Same issue — issue_comment also signs without enqueuing**

Same pattern as publish_discussion_comment: the old send_event_builder call was replaced with sign_event_builder, but the publish_queue::enqueue() step was never added. The signed event is returned but never sent to any relay.

These two functions (publish_discussion_comment and publish_issue_comment) appear to be the only git hosting publish functions that lost their relay send step during the migration to the publish queue.

src/stores/cashu/init.rs
```diff
@@ -210,23 +210,18 @@ pub async fn accept_terms() -> Result<(), String> {
...
-    if output.success.is_empty() {
-        return Err("Failed to publish terms acceptance: no relay accepted".to_string());
-    }
-    log::info!("Terms acceptance published successfully");
+    crate::stores::publish_queue::enqueue(
+        event,
+        crate::stores::publish_queue::types::QueueEventType::Cashu,
+        None,
+        std::collections::HashMap::new(),
+    ).await;
```
goodmorningbitcoin: **Cashu accept_terms() sets TERMS_ACCEPTED before confirming the event was enqueued**

The function enqueues the terms acceptance event and then immediately sets TERMS_ACCEPTED to Some(true), returning Ok(()). But enqueue() is fire-and-forget — it does not guarantee the event reaches any relay.

If the queue processing fails (e.g., no relays connected, all relays reject), the app will think terms were accepted when they were not. On the next app launch, TERMS_ACCEPTED would be reset but the user may have already been shown the wallet UI.

This is lower severity than create_wallet() (which is the critical one), but the same pattern applies. Consider enqueue_and_await() for operations where the success/failure matters before updating local state.

src/utils/nips/nip36.rs
patrickulrich: nip36.rs Option<Option<String>> — NOT a bug
The Option<Option<String>> return type is semantically correct and maps directly to NIP-36's spec:
- None = no content-warning tag exists
- Some(None) = tag exists, no reason text
- Some(Some("reason")) = tag exists with reason
All 13 callers use if let Some(reason) = content_warning which destructures to reason: Option<String> — exactly matching SensitiveContent's reason: Option<String> prop. The behavior is correct in all three cases. This is a style concern, not a bug.

src/routes/events/calendar_event_new.rs
```diff
@@ -66,11 +68,150 @@ pub fn CalendarEventNew() -> Element {
...
+                        timezone.set(draft.timezone);
+                        participants.set(draft.participants);
+                    }
+                }
+            }
+        }
+    });
+    // Draft auto-save (only when NOT in edit mode)
+    use_effect(move || {
+        if *is_edit_mode.read() {
```
goodmorningbitcoin, line 183: **Calendar draft auto-save fires on every keystroke with no debounce**

Same issue as the note composer draft from the earlier review. The use_effect in calendar_event_new.rs reads every signal (title, summary, content, start_date, start_time, end_date, end_time, location, locations, image_url, hashtags_input, timezone, participants) and writes to storage on every change.

Each keystroke in any field triggers a full draft serialization + storage write. On web this is localStorage.setItem() (synchronous main-thread block); on native it is a file write behind a Mutex.

Recommend adding a debounce (e.g., 500ms-1000ms delay before writing) the same way the note composer draft should be debounced.

Ref: Same pattern as the earlier draft-auto-save comment on use_composer_editor.rs.

src/stores/publish_queue/persistence.rs
```diff
@@ -0,0 +1,161 @@
...
+                vec![]
+            }),
+            Err(_) => vec![],
+        }
+    }
+
+    fn save_queue(events: &[QueuedEvent]) -> Result<(), String> {
+        let path = queue_path();
+        if let Some(parent) = path.parent() {
+            std::fs::create_dir_all(parent).map_err(|e| e.to_string())?;
```
goodmorningbitcoin, line 80: **Native persistence functions are unnecessarily async — std::sync::Mutex held across sync-only code**

All the native persistence functions (add_queued_event, get_all_queued_events, remove_queued_event, update_queued_event) are declared async but perform only synchronous file I/O. The std::sync::Mutex is acquired at the top and released when the function returns.

This is functionally correct since the actual work is sync (no await points inside the lock), but:
1. The async signature is misleading — callers .await these functions for no reason
2. The Mutex<()> guards file access, which is fine, but every operation does a full load-modify-save cycle (read entire JSON, modify, write entire JSON). Under high enqueue throughput this is a lot of redundant serialization.

Consider making these sync functions and having callers use spawn_blocking() if needed, or at minimum document that they are intentionally sync-in-async clothing.

This is low severity — the current code works correctly.

src/routes/events/event_detail.rs
```diff
@@ -363,25 +369,209 @@ pub fn CalendarEventDetail(naddr: String, from: Option<String>) -> Element {
...
+                                                                                    Ok(result) => {
+                                                                                        if result.is_success() {
+                                                                                            toast_api.success(
+                                                                                                "Broadcast queued".to_string(),
+                                                                                                ToastOptions::new().duration(Duration::from_secs(3)).permanent(false),
+                                                                                            );
+                                                                                        } else {
+                                                                                            toast_api.error(
+                                                                                                "Broadcast failed".to_string(),
```
goodmorningbitcoin, line 479: **Broadcast fallback uses EventId::all_zeros() when hex parse fails — silent data corruption**

If EventId::from_hex(&eid) fails, the code falls back to EventId::all_zeros() and fetches that from relays. This will either return nothing or return a wrong event, then attempt to broadcast it. The error is silently swallowed.

Should error out or show a toast instead of using a sentinel value.

Also: the broadcast flow refetches the event from relays by ID even though the event data is already available locally in the UnifiedEvent. This is an unnecessary network round-trip — the signed event could be reconstructed or cached.

src/routes/events/calendar_event_new.rs
```diff
@@ -361,6 +513,7 @@ pub fn CalendarEventNew() -> Element {
                         &all_locations,
                         &hashtags,
                         &participants_val,
+                        edit_d_tag_val.as_deref(),
```
goodmorningbitcoin, line 516: **Calendar event edit allows changing type (DateBased <-> TimeBased) which creates a new event instead of replacing**

When editing a calendar event, the code passes existing_d_tag through to publish_date_event/publish_time_event. For NIP-52 addressable events, relays use (kind, author, d_tag) as the unique key. If the user changes the event type during edit, the kind changes (31922 vs 31923) while the d_tag stays the same — this creates a second event rather than replacing the original.

The edit UI should either lock the event type selector in edit mode, or show a warning that changing type creates a new event rather than updating the existing one.

src/routes/events/event_detail.rs
```diff
@@ -363,25 +369,209 @@ pub fn CalendarEventDetail(naddr: String, from: Option<String>) -> Element {
...
+                                                            }
+                                                            toast_api.success(
+                                                                "Copied to clipboard".to_string(),
+                                                                ToastOptions::new().duration(Duration::from_secs(2)).permanent(false),
+                                                            );
+                                                        });
+                                                        is_menu_open.set(false);
+                                                    },
+                                                    if *is_copying.read() { "Copying..." } else { "Copy Event ID" }
+                                                }
```
goodmorningbitcoin, line 542: **Copy Event ID on non-web does nothing, but still shows success toast**

The non-web branch has `let _ = naddr_str;` (no-op) but still calls toast_api.success("Copied to clipboard"). On native/mobile, the user sees a success message but nothing was actually copied.

Either implement clipboard for native (e.g., arboard crate or platform API) or show a different message / disable the button on non-web.

PR #298, opened 2026-04-29
Release 0.8.8: AI settings sync, AAB wallet fix, CI fix

patrickulrich:
- **Sync AI provider credentials across devices** — Persist AI provider settings to Nostr relays via encrypted NIP-44 events, enabling cross-device credential sync with automatic retry for failed states
- **Exclude Cashu wallet from Play Store AAB** — Move `cdk-sqlite` dependency out of the shared `native` feature and into `desktop`/`mobile` only, so the Play Store build no longer bundles the CDK dependency tree
- **Add version-based changelog files for CI** — CI now reads `public/docs/changelogs/{version}.md` for GitHub release notes, falling back to "Automated release from main." when no file exists
- **Center app icon** — Fix vertically off-center N glyph in Android adaptive icon foreground assets and main icon PNGs


src/stores/ui/ai_provider_store.rs
```diff
@@ -319,20 +439,33 @@ mod web_db {
...
-        if let Ok(cached_state) = storage::get(STORAGE_KEY) {
-            return Ok(migrate_legacy_state(cached_state));
+    let mut state = {
+        #[cfg(all(target_arch = "wasm32", feature = "web", not(feature = "native")))]
+        {
+            if let Ok(cached_state) = storage::get(STORAGE_KEY) {
+                migrate_legacy_state(cached_state)
+            } else {
+                web_db::AiProviderDb::new().await?.load_state().await?
+            }
```
goodmorningbitcoin, line 449: **Relay state unconditionally overwrites local state — no merge, no timestamp comparison**

In load_provider_state(), when pending_relay_state exists, it does state = relay_state — a full replacement of whatever was loaded from local storage. If device A has newer credentials than device B, and device B logs in and fetches the relay event (which was published by device A before device B made local changes), device B's local changes are silently lost.

Consider: (1) storing a timestamp or version counter in the encrypted payload and taking the latest, or (2) merging custom_providers by ID (union, preferring the relay version on conflict). At minimum, document that relay wins unconditionally.

src/stores/ui/ai_provider_store.rs
```diff
@@ -225,6 +341,10 @@ pub fn process_queued_provider_state_saves(initial_snapshot: AiProviderState) {
         let mut next_snapshot = Some(initial_snapshot);
         while let Some(current_snapshot) = next_snapshot {
             let result = save_provider_state(&current_snapshot).await;
+            let encrypted_result = save_encrypted_provider_state(&current_snapshot).await;
```
goodmorningbitcoin, line 344: **Encrypted save is fire-and-forget on error — local save succeeds but relay sync silently fails**

process_queued_provider_state_saves() calls save_provider_state() (local) then save_encrypted_provider_state() (relay). If the encrypted save fails, it only logs a warning — the local save is considered successful and the emit_provider_state_save_event reports success.

This means the user sees a success notification but their credentials are not synced to relays. On next login from another device, the stale relay state overwrites their local state (see the load_provider_state comment above).

Consider retrying the encrypted save, or surfacing a discrete warning that sync failed.

src/stores/nostr_client/mod.rs
```diff
@@ -298,12 +298,19 @@ pub async fn initialize_client() -> std::result::Result<Arc<Client>, String> {
...
-            log::info!("Background relay connections completed");
+            let output = client_for_connect.try_connect(Duration::from_secs(3)).await;
+            let success = !output.success.is_empty();
+            let _ = tx.send(success);
+            log::info!("Background relay connections completed (success={})", success);
+        });
+        spawn_forever(async move {
+            if let Some(success) = rx.recv().await {
+                if success && !*RELAY_CONNECTED.peek() {
+                    *RELAY_CONNECTED.write() = true;
```
goodmorningbitcoin, line 311: **spawn_forever in initialize_client silently dies after one message**

The non-wasm relay connection path uses spawn_forever to receive the result from the tokio::spawn task. spawn_forever runs an async future, but the body does rx.recv().await once and then the future completes. If RELAY_CONNECTED is already true at that point, the future just ends — which is fine.

But the name spawn_forever implies the task persists. This task does not — it receives one message and exits. The spawn_forever is just being used as a Dioxus-compatible way to get back onto the UI thread for writing RELAY_CONNECTED. This works, but is slightly misleading. A comment explaining the intent would help future readers.

src/main.rs
```diff
@@ -127,6 +127,23 @@ fn App() -> Element {
...
         }
     });
+    use_effect(move || {
+        let is_authenticated = auth_store::AUTH_STATE.read().is_authenticated;
+        if is_authenticated {
+            let sidebar_failed = sidebar_store::SIDEBAR_STATE.peek().is_failed();
+            let reactions_failed = reactions_store::REACTIONS_STATE.peek().is_failed();
+            if sidebar_failed || reactions_failed {
+                spawn(async move {
+                    if sidebar_store::SIDEBAR_STATE.peek().is_failed() {
```
goodmorningbitcoin, line 137: **New use_effect retries on every render while sidebar/reactions remain failed**\n\nThe first retry block (line 115) correctly triggers only when RELAY_CONNECTED changes — that is reactive and fine.\n\nThis new block (line 137) triggers on every render when authenticated + failed. Since the load functions are async, and the state may remain is_failed() until they complete, this will fire on every Dioxus re-render cycle, spawning multiple concurrent loads.\n\nConsider either: (1) adding an in_flight guard so only one retry runs at a time, or (2) removing this block and relying on the relay-connected handler above.

.github/workflows/deploy.yml
```diff
@@ -405,6 +405,21 @@ jobs:
...
+        run: |
+          FILE="public/docs/changelogs/${VERSION}.md"
+          if [ -f "$FILE" ]; then
+            echo "Using changelog from $FILE"
+            {
+              echo 'RELEASE_NOTES<<EOF'
+              cat "$FILE"
+              echo ''
+              echo 'EOF'
+            } >> "$GITHUB_ENV"
```
goodmorningbitcoin, line 420: **RELEASE_NOTES may be unset when gh release create runs**

If no changelog file exists for the version, RELEASE_NOTES is never set in the environment. The gh release create command then runs with --notes "" (empty string) instead of falling back to the PR description body that was in the original code.

Original behavior had --notes "Automated release from main." as a hardcoded fallback. Now there is no fallback — you get an empty release body. Consider adding an else branch:

if [ -f "$FILE" ]; then
  ...
else
  echo 'RELEASE_NOTES=Automated release from main.' >> "$GITHUB_ENV"
fi

patrickulrich: Yes. The fallback is on line 21 of .github/workflows/deploy.yml:
RELEASE_NOTES: "Automated release from main."
That's set as a workflow-level env var. The "Resolve release notes" step we added only overrides it when the file exists — otherwise the default message is used unchanged.

PR #306, opened 2026-04-30
v0.8.9: NIP-39 external identities, copy note, embedded note rendering, and mobile fixes

patrickulrich:
- **NIP-39 External Identity Verification** — Display verified identity badges (GitHub, X/Twitter, Mastodon, Telegram) on profile pages, with profile editor fields for linking accounts via proof URLs. Parses kind 10011 events using SDK's `TagStandard::ExternalIdentity`.
- **Copy Note** — Add "Copy Note" menu item to note context menu with native Android clipboard support via JNI bridge.
- **Embedded Note Rendering** — Fix embedded/quoted notes rendering as plain text; now uses `RichContent` for proper formatting.
- **Mention Autocomplete** — Fix `@` mention autocomplete not working on mobile devices.
- **NIP-55 Android Signer** — Fix mismatched package false positive when detecting NIP-55 signer apps on Android.


goodmorningbitcoin (commented):
## Code Review Summary

Reviewed all 18 files in this PR. Found several issues ranging from a compile-breaking duplicate to UX/logic bugs.

### Critical (1)
- **Duplicate `MastodonIcon` definition** — will fail to compile since the icon already exists at line 1045 of `icons.rs`

### High (4)
- Embedded note `onclick` is a no-op on native/mobile builds — the `navigator().push()` call is inside `#[cfg(feature = "web")]`
- `display_name()` for Mastodon returns the server domain, not the username — users see "bitcoinhackers.org" instead of "@semisol"
- `cfg!(feature = "web")` in `render_dropdown` keeps dead code in the binary — use `#[cfg]` attributes instead
- Profile editor doesn't allow removing/ clearing existing identity proofs (no way to delete, only overwrite)

### Medium (4)
- Copy Note button has no icon (inconsistent with the UI pattern of other menu items, though I see some items also lack icons)
- `get_cursor_position` non-web fallback returns end-of-string position — misleading for the mention detection logic
- No Telegram proof URL field in the profile editor (Telegram identities can be fetched but can't be created/edited)
- `publish_external_identities` replaces all identities atomically — if the user only edits one field, unchanged identities from other platforms must also be re-published or they'll be lost

### Low (2)
- `KIND_EXTERNAL_IDENTITIES` is `u16` but should use `Kind::Custom` constant or be properly typed
- Clipboard read on Android returns raw clipboard content with no sanitization

src/components/rich_content/mentions.rs
```diff
@@ -378,37 +382,71 @@ pub(super) fn render_embedded_note(event: &Event, metadata: Option<&Metadata>) -
...
-                from_voice: None,
-            },
-            class: "block my-2",
-            onclick: move |e: MouseEvent| e.stop_propagation(),
-            div { class: "bg-card border border-border rounded-lg p-4 hover:bg-accent/10 transition cursor-pointer",
-                div { class: "flex items-center gap-2 mb-2",
-                    if let Some(meta) = metadata {
-                        if let Some(picture) = meta.picture.as_ref().filter(|u| is_valid_http_url(u)) {
-                            img {
-                                class: "w-8 h-8 rounded-full",
```
goodmorningbitcoin, line 393: **Bug: onclick is a no-op on native/mobile builds**

The entire interactive-element click guard is wrapped in `#[cfg(feature = "web")]`. On native (desktop/mobile) builds, the `onclick` handler falls through to nothing. The embedded note is completely unclickable on Android and desktop.

The fix: Move the `navigator().push()` call outside the `#[cfg(feature = "web")]` block, and only guard the `closest()` DOM check for web. Same issue applies to the `onkeydown` handler above.

src/utils/nips/nip39.rs
```diff
@@ -0,0 +1,296 @@
...
+
+    pub fn display_name(&self) -> &str {
+        if self.platform == "mastodon" {
+            if let Some(pos) = self.ident.find("/@") {
+                return &self.ident[..pos];
+            }
+        }
+        &self.ident
+    }
+}
```
goodmorningbitcoin, line 44: **Bug: `display_name()` for Mastodon returns the server domain instead of the username**

For a Mastodon identity with `ident = "bitcoinhackers.org/@semisol"`, this returns `"bitcoinhackers.org"` (the server). Users would expect to see `"@semisol"` or `"semisol@bitcoinhackers.org"`.

The `find("/@")` locates the `/@` separator and returns everything before it — which is the domain.

```rust
pub fn display_name(&self) -> &str {
    if self.platform == "mastodon" {
        if let Some(pos) = self.ident.find("/@") {
            // self.ident = "bitcoinhackers.org/@semisol"
            // pos = 18 (index of '/')
            // returns "bitcoinhackers.org" — the SERVER, not the user
            return &self.ident[..pos];
        }
    }
    &self.ident
}
```

**Fix:** Return the username portion after `/@`:
```rust
if let Some(pos) = self.ident.find("/@") {
    return &self.ident[pos + 2..]; // "semisol"
}
```

Or for a more traditional Mastodon format: `"semisol@bitcoinhackers.org"`.

src/components/mention_autocomplete.rs
```diff
@@ -81,29 +84,55 @@ pub fn MentionAutocomplete(props: MentionAutocompleteProps) -> Element {
...
+                cursor_pos,
+                &mut autocomplete,
+                contact_pubkeys,
+                &props.thread_participants,
             );
+            if *autocomplete.show.read() {
+                update_dropdown_position(
+                    &textarea_id.read(),
+                    &mut dropdown_top,
+                    &mut dropdown_left,
```
goodmorningbitcoin, line 109: **`cfg!(feature = "web")` in runtime `class` bindings keeps dead code in the binary**

Using `cfg!(...)` as a runtime boolean means both branches compile and end up in the WASM/native binary. Since the class strings are compile-time constants, prefer `#[cfg]`-gated bindings:

```rust
// Instead of:
class: if cfg!(feature = "web") { "fixed ..." } else { "absolute ..." }

// Use:
#[cfg(feature = "web")]
class: "fixed bg-white dark:bg-gray-800 shadow-lg rounded-lg border ...",
#[cfg(not(feature = "web"))]
class: "absolute top-full left-0 right-0 mt-1 bg-white dark:bg-gray-800 ...",
```

However, Dioxus RSX macros may not support multiple `class` attributes with cfg guards. If that's the case, extracting the class into a helper function with `#[cfg]` would be cleaner and produce smaller binaries.

src/components/rich_content/mentions.rs
patrickulrich: This bug report is inaccurate. Looking at the actual code in render_embedded_note (lines 358-453):
onclick handler (lines 409-424):
onclick: move |_evt: MouseEvent| {
    #[cfg(feature = "web")]
    {
        // interactive element guard - only on web
    }
    // This is OUTSIDE the cfg block:
    navigator().push(Route::Note { ... });
},
The navigator().push() call is already outside the #[cfg(feature = "web")] block. On native/desktop/mobile builds, the cfg-guarded block is simply compiled out, and navigator().push() executes unconditionally. The same pattern applies to the onkeydown handler (lines 390-408).
No fix needed — the code is correctly structured.

PR #309, opened 2026-05-05
Release 0.8.10: fix comment threading, music player perf, cross-platform Blossom, Dioxus 0.7.7

patrickulrich:
## Performance
- Fix Android UI freeze from music player re-rendering entire state every 1s — switch to field-level Store selectors and extract seek bar into child component to isolate time-tick re-renders
- Parallel prefetch for notification author profiles and referenced posts
## Bug Fixes
- Fix comment thread nesting — `get_parent_id()` was using broken `tag.content().split('\t')` marker parsing that never detected NIP-10 Reply/Root markers, causing all replies to comments to appear at the top level. Rewritten using `tag.as_standardized()` with proper `TagStandard::Event` matching
- Fix reply cache invalidation using wrong event ID as key
- Port Blossom list/delete/mirror from wasm-only `gloo_net` to `reqwest` for cross-platform support (desktop, mobile, playstore)
## Dependencies
- Upgrade Dioxus 0.7.5 → 0.7.7


goodmorningbitcoin (commented):
Review in progress

goodmorningbitcoin:
## 🔍 Code Review: PR #309

### Critical Issues 🐛

**1. Broken seek calculation in `player_bar.rs` (line 150)**
```javascript
// BEFORE (correct):
let percent = Math.max(0, Math.min(1, ({client_x} - rect.left) / rect.width));

// AFTER (broken):
let percent = Math.max(0, Math.min(1, ({client_x} - rect.width));
```
The `- rect.left` was removed along with the `/ rect.width` division. This means:
- Division by `rect.width` is gone — seek position will be wildly wrong
- `client_x - rect.width` makes no sense as a seek calculation
This will cause the seek bar on mobile to compute completely incorrect positions. **Must fix before merge.**

---

### Significant Concerns ⚠️

**2. Timeout removed from Blossom HTTP requests** (`blossom_store.rs`)
All three Blossom endpoints (`fetch_server_list`, `delete_from_server`, `mirror_to_server`) lost their 10-second timeout. Unresponsive servers will hang indefinitely. The old `AbortController` + `Timeout` pattern should be replicated in the new `http_client()`-based code.

**3. Race condition in `show_zap_dialog_for_track`** (`music_player.rs` line 868)
`store.current_track().peek().clone()` reads without a lock, then `store.zap_track().write()` is a separate write. The old `MUSIC_PLAYER.write()` made these atomic. With field-level selectors, there's no global lock. The race window is tiny but worth documenting.

**4. Behavioral change in `sync_native_playback_snapshot`** (`music_player.rs` line 1136)
Now skips syncing `current_track` when index hasn't changed. If playlist entries mutate (e.g., `media_url` changes during fallback), the old code would update `current_track` — the new code won't. Verify this is intentional.

**5. Floating player position divergence** (`player_floating.rs` line 9)
`pos` signal is initialized from store on mount but never re-synced. If `set_floating_pos` is called externally, the floating player won't move.

---

### Observations 🤔

**6. Thread tree `get_parent_id` semantic change** (`thread_tree.rs` line 72)
Falls back to `Marker::Root` as parent when no `Marker::Reply` exists. Old code returned the **last** `e` tag. New code returns the first `e` tag with `marker=Root`. This changes thread tree construction — make sure it matches your NIP-10 interpretation.

**7. Infinite scroll polling 250ms → 1000ms** (`use_infinite_scroll.rs` line 325)
4x slower sentinel detection. Users may notice visible delay when scrolling quickly. Consider a middle ground or backoff.

**8. Notification prefetch fetches all `e` tags** (`notifications.rs` line 1575)
Not just parent/root — all event references. Minor inefficiency, consider filtering by marker.

---

### Positive Notes 👍
- Extracting `ExpandedSeekBar` as a child component with primitive `f64` props is a solid perf pattern — avoids re-rendering the entire expanded player on every time tick
- The `reply_composer.rs` cache invalidation fix (using actual `root` reference) is a genuine bugfix — nice catch
- Cleaning up the 2672-line `home.rs.bak` is good hygiene
- Blossom store refactor removing `gloo_net` for `http_client` is clean cross-platform support
- Dioxus 0.7.7 upgrade looks straightforward

---

### Verdict
**Request changes** — the seek bar JS bug needs fixing before merge. The timeout removal on Blossom requests and the behavioral changes in `sync_native_playback_snapshot` / thread tree logic should also be addressed or confirmed intentional.

src/components/music/player/player_bar.rs
```diff
@@ -139,7 +147,7 @@ pub fn PlayerBar() -> Element {
                             if (!element) return null;
                             let progressBar = element.closest('.cursor-pointer') || element;
                             let rect = progressBar.getBoundingClientRect();
-                            let percent = Math.max(0, Math.min(1, ({client_x} - rect.left) / rect.width));
+                            let percent = Math.max(0, Math.min(1, ({client_x} - rect.width));
```
goodmorningbitcoin, line 150: 🐛 **Critical Bug: Missing division in seek calculation**

```javascript
// BEFORE (correct):
let percent = Math.max(0, Math.min(1, ({client_x} - rect.left) / rect.width));

// AFTER (broken):
let percent = Math.max(0, Math.min(1, ({client_x} - rect.width));
```

The `- rect.left` was removed along with the `/ rect.width` division. This means:
1. The division by `rect.width` is gone — seek position will be wildly wrong
2. `client_x - rect.width` makes no sense as a seek calculation (it should be `client_x - rect.left` divided by `rect.width`)

This will cause the seek bar on the mobile platform player to compute completely incorrect positions. **Must fix before merge.**

src/stores/audio/music_player.rs
```diff
@@ -843,32 +843,32 @@ pub fn close_player() {
...
+    let store = MUSIC_PLAYER.resolve();
+    *store.zap_track().write() = track.or_else(|| store.current_track().peek().clone());
+    *store.show_zap_dialog().write() = true;
 }
 /// Show zap dialog for current track
 pub fn show_zap_dialog() {
     show_zap_dialog_for_track(None);
 }
 /// Hide zap dialog
 pub fn hide_zap_dialog() {
```
goodmorningbitcoin, line 868: ⚠️ **Race condition: `peek()` + `or_else(clone)` across separate reads**

```rust
let store = MUSIC_PLAYER.resolve();
*store.zap_track().write() = track.or_else(|| store.current_track().peek().clone());
*store.show_zap_dialog().write() = true;
```

`store.current_track().peek()` reads without subscribing, and `store.zap_track().write()` is a separate write. In the old code with `MUSIC_PLAYER.write()`, both reads and writes happened under a single write lock, making them atomic.

With field-level selectors, there is no global lock. Consider whether this non-atomicity matters for your use case. If `show_zap_dialog_for_track(None)` is always called when a track is playing, the race window is tiny, but worth documenting.

src/stores/audio/music_player.rs
```diff
@@ -1120,41 +1121,43 @@ pub fn set_available_streams(streams: Vec<String>) {
...
 pub fn sync_native_playback_snapshot(snapshot: AndroidPlaybackSnapshot) {
-    let has_error = snapshot.playback_error.is_some();
-    let mut state = MUSIC_PLAYER.write();
-    if !state.playlist.is_empty() && snapshot.current_index < state.playlist.len() {
-        state.current_index = snapshot.current_index;
-        state.current_track = state.playlist.get(snapshot.current_index).cloned();
+    let store = MUSIC_PLAYER.resolve();
+    let prev_index = *store.current_index().peek();
+    if !store.playlist().peek().is_empty() && snapshot.current_index < store.playlist().peek().len()
+        && prev_index != snapshot.current_index {
```
goodmorningbitcoin, line 1136: 🤔 **Behavioral change — now skips index sync when unchanged**

```rust
// NEW: only syncs index if it changed
if !store.playlist().peek().is_empty() && snapshot.current_index < store.playlist().peek().len()
    && prev_index != snapshot.current_index {
    *store.current_index().write() = snapshot.current_index;
    *store.current_track().write() = ...
}
```

The old code always set `current_index` and `current_track`, even if the index was the same. The new code skips the update when `prev_index == snapshot.current_index`. This is a behavioral change — if the track object in the playlist was replaced (e.g., URL changed for fallback), the old code would update `current_track` to match. The new code won't.

Is this intentional? If the playlist entries can mutate (e.g., `media_url` changes during fallback), this could cause stale track data.

src/stores/media/blossom_store.rs
```diff
@@ -814,44 +812,22 @@ async fn list_files_inner() -> Result<Vec<MediaItem>, String> {
...
+        .await
+        .map_err(|e| format!("Request failed: {}", e))?;
+    if !response.status().is_success() {
+        return Err(format!("Server returned status: {}", response.status()));
     }
+    let blobs: Vec<BlobDescriptor> = response
+        .json()
+        .await
+        .map_err(|e| format!("Failed to parse response: {}", e))?;
+    Ok(blobs)
```
goodmorningbitcoin, line 830: ⚠️ **Timeout removed from HTTP requests**

The old code used `AbortController` + `Timeout` to enforce a 10-second timeout on all Blossom server requests. The new `http_client()`-based code has no timeout.

If a Blossom server is unresponsive, requests will hang indefinitely. Consider adding a timeout to the `http_client()` builder or wrapping calls with a platform-appropriate timeout mechanism.

This affects `fetch_server_list`, `delete_from_server`, and `mirror_to_server` — all three lost their timeout protection.

src/utils/parsing/thread_tree.rs
```diff
@@ -33,45 +34,46 @@ impl ThreadNode {
...
             }
         }
     }
+
+    if reply_marker_id.is_some() {
+        return reply_marker_id;
+    }
+    if root_marker_id.is_some() {
+        return root_marker_id;
+    }
```
goodmorningbitcoin, line 72: 🤔 **Thread tree `get_parent_id` logic change — Root used as parent**

```rust
if reply_marker_id.is_some() {
    return reply_marker_id;
}
if root_marker_id.is_some() {
    return root_marker_id;  // Root marker used as parent?
}
```

When there is no `reply` marker but there IS a `root` marker, this returns the `root` event ID as the parent. Per NIP-10, a root marker indicates the thread root, not necessarily the direct parent.

**Old code**: With a single `e` tag, returned that tag's event ID. With multiple `e` tags and no explicit `reply` marker, returned the **last** `e` tag.

**New code**: Returns the **first** `e` tag with `marker=Root` as the parent when no Reply marker exists.

Consider a case with 3 e-tags: `[e(root), e(mention), e(mention)]` — old code returns the last mention, new code returns root. This changes thread tree construction. Make sure this matches your intended NIP-10 interpretation.

src/hooks/use_infinite_scroll.rs
```diff
@@ -322,7 +322,7 @@ where
                         }
                     }
 
-                    crate::platform::timer::sleep_ms(250).await;
+                    crate::platform::timer::sleep_ms(1000).await;
```
goodmorningbitcoin, line 325: 🤔 **Polling interval changed from 250ms to 1000ms**

```rust
// Before:
crate::platform::timer::sleep_ms(250).await;
// After:
crate::platform::timer::sleep_ms(1000).await;
```

This is a 4x slowdown in infinite scroll polling. The scroll sentinel detection will now have up to 1 second of latency before triggering the next page load.

Was this intentional? If it is for performance, consider that users on fast connections will notice a visible delay when scrolling quickly through content. A middle ground (500ms) or backoff-based approach might be better.

src/components/music/player/player_floating.rs
```diff
@@ -1,14 +1,16 @@
...
+use crate::stores::music_player::{self, MusicPlayerStateStoreExt, MUSIC_PLAYER};
 use dioxus::prelude::*;
 
 #[component]
 pub fn PlayerFloating() -> Element {
-    let state = MUSIC_PLAYER.read().clone();
-    let track = state.current_track.as_ref().unwrap();
+    let store = MUSIC_PLAYER.resolve();
+    let track = store.current_track().cloned().unwrap();
+    let floating_pos = store.floating_pos().cloned();
```
goodmorningbitcoin, line 9: ⚠️ **Floating position initialized once, may drift from store**

```rust
let floating_pos = store.floating_pos().cloned();
let mut pos = use_signal(|| floating_pos);
```

`pos` is initialized from `floating_pos` on mount, but never re-syncs if the store's `floating_pos` changes externally (e.g., another call to `set_floating_pos`). The old code cloned the full state each render, so `use_signal(|| state.floating_pos)` would pick up changes on re-render.

Now, `pos` is a local signal that diverges from the store after mount. If `set_floating_pos` is called from outside this component, the floating player won't move. This may be intentional (local drag state), but worth verifying.

src/components/music/player/player_bar.rs
```diff
@@ -1,17 +1,27 @@
...
-    let state = MUSIC_PLAYER.read().clone();
-    let track = state.current_track.as_ref().unwrap();
-    let progress = if state.duration > 0.0 {
-        (state.current_time / state.duration * 100.0).clamp(0.0, 100.0)
+    let store = MUSIC_PLAYER.resolve();
+    let track = store.current_track().cloned().unwrap();
+    let current_time = store.current_time().cloned();
+    let duration = store.duration().cloned();
+    let is_playing = store.is_playing().cloned();
+    let is_buffering = store.is_buffering().cloned();
```
goodmorningbitcoin, line 16: 💡 **Extra `.cloned()` calls on primitive fields**

```rust
let current_time = store.current_time().cloned();
let duration = store.duration().cloned();
let is_playing = store.is_playing().cloned();
```

These are all `f64` or `bool` — `Copy` types. Using `.cloned()` works, but `*store.field().read()` is more idiomatic for `Copy` types. The `player_expanded.rs` file already does this:

```rust
let current_time = *store.current_time().read();
let duration = *store.duration().read();
```

Consider using the deref pattern consistently across all player components. Minor nitpick.

src/components/music/player/expanded_seek_bar.rs
```diff
@@ -0,0 +1,128 @@
...
+#[component]
+pub fn ExpandedSeekBar(current_time: f64, duration: f64) -> Element {
+    let mut is_scrubbing = use_signal(|| false);
+    let mut scrub_position = use_signal(|| None::<f64>);
+    let mut seek_bar_left = use_signal(|| 0.0f64);
+    let mut seek_bar_width = use_signal(|| 1.0f64);
+    let mut gesture_id = use_signal(|| 0u32);
+
+    let progress = if duration > 0.0 {
+        (current_time / duration * 100.0).clamp(0.0, 100.0)
```
goodmorningbitcoin, line 15: 👍 **Seek bar as isolated child component — good perf pattern**

Extracting the seek bar into `ExpandedSeekBar` with `current_time: f64` and `duration: f64` as props means Dioxus can skip re-rendering the parent `PlayerExpanded` on every time tick. Time updates only flow into this child component.

This is a solid pattern for reducing unnecessary re-renders in a music player. The new file is clean and well-structured.

src/components/reply_composer.rs
```diff
@@ -158,8 +158,10 @@ pub fn ReplyComposer(
             log::info!("Enqueued reply: {}", event_id);
             show_queued_toast(toast_api, "Reply");
 
-            if let Ok(root_event_id) = EventId::from_hex(&target_event.id.to_hex()) {
-                invalidate_thread_tree_cache(&root_event_id);
+            if let Some(ref r) = root {
```
goodmorningbitcoin, line 161: ✅ **Cache invalidation fix: correctly uses `root` reference**

Old code invalidated cache with the **target** event ID (the one being replied to), not the thread root. If replying to a mid-thread comment, only that sub-tree would be invalidated.

New code correctly uses the actual `root` reference if available, falling back to `target_event.id`. This is a genuine bugfix — nice catch.

src/routes/notifications.rs
```diff
@@ -1513,3 +1555,45 @@ async fn prefetch_notification_authors(notifications: &[NotificationType]) {
...
+        let event = match notif {
+            NotificationType::Reaction(e)
+            | NotificationType::Repost(e)
+            | NotificationType::Quote(e)
+            | NotificationType::Zap(e) => e,
+            _ => continue,
+        };
+        for tag in event.tags.iter() {
+            if let Some(nostr_sdk::TagStandard::Event { event_id, .. }) = tag.as_standardized() {
+                if seen.insert(*event_id) {
```
goodmorningbitcoin, line 1575: ⚠️ **`prefetch_notification_posts` fetches all `e` tags, not just parent/root**

```rust
for tag in event.tags.iter() {
    if let Some(nostr_sdk::TagStandard::Event { event_id, .. }) = tag.as_standardized() {
        if seen.insert(*event_id) {
            event_ids.push(*event_id);
        }
    }
}
```

This collects **all** `e` tags from notification events, not just the parent/root references. A reaction event might reference the reacted-to post AND the thread root — both get prefetched. While harmless (extra caching), it means potentially 2x the fetches needed.

Also: the `.truncate(100)` limit is good, but consider logging when truncation occurs so you can monitor if this cap is hit in practice.

src/stores/audio/music_player.rs
```diff
@@ -1074,38 +1074,39 @@ pub fn try_next_stream() -> bool {
...
+    *store.is_buffering().write() = true;
+    if let Some(ref mut track) = store.current_track().write().as_mut() {
         track.media_url = new_url;
     }
-    let playlist = state.playlist.clone();
-    let index = state.current_index;
-    drop(state);
+    let playlist = store.playlist().peek().clone();
+    let index = *store.current_index().peek();
     log::info!(
```
goodmorningbitcoin, line 1100: ⚠️ **Redundant `MUSIC_PLAYER.resolve()` call — creates second resolver**

```rust
if let Err(e) = crate::platform::android_media::set_queue(&playlist, index, true) {
    log::error!("Failed to start Android fallback stream: {}", e);
    let store = MUSIC_PLAYER.resolve();  // ← second resolve()
    *store.playback_error().write() = Some(format!("Fallback stream failed: {}", e));
    *store.is_buffering().write() = false;
    return false;
};
```

This creates a **second** `store` binding. The first `store` from earlier in the function is still in scope. While this works (each `.resolve()` returns a new handle), it is wasteful and inconsistent. Could just reuse the original `store` binding. Not a bug, but unnecessary allocation.

src/routes/notifications.rs
```diff
@@ -231,7 +232,10 @@ pub fn Notifications() -> Element {
                         oldest_timestamp.set(oldest);
                         has_more.set(count >= 100);
                         spawn(async move {
-                            prefetch_notification_authors(&notifs).await;
+                            join!(
```
goodmorningbitcoin, line 235: 💡 **Pattern: identical `join!` block repeated 3 times**

The same `join!(prefetch_notification_authors(&notifs), prefetch_notification_posts(&notifs))` pattern appears in three separate `spawn` blocks (lines 235, 285, 338 — initial load, load more, and gap fill). Consider extracting a helper:

```rust
async fn prefetch_notification_data(notifs: &[NotificationType]) {
    join!(
        prefetch_notification_authors(notifs),
        prefetch_notification_posts(notifs),
    );
}
```

Reduces duplication and makes it easier to add more prefetch logic later.

src/utils/parsing/thread_tree.rs
```diff
@@ -33,45 +34,46 @@ impl ThreadNode {
...
-                if let Ok(event_id) = EventId::from_hex(parts[0]) {
-                    return Some(event_id);
+                None => {
+                    last_unmarked_id = Some(*event_id);
                 }
             }
         }
     }
+
+    if reply_marker_id.is_some() {
```
goodmorningbitcoin, line 67: 💡 **Minor: early-return pattern could be simplified**

Since you guard with `if reply_marker_id.is_none()` during assignment, the end-of-function could be more idiomatically written as:

```rust
reply_marker_id.or(root_marker_id).or(last_unmarked_id)
```

No change required — the current code is clear and defensive. Just an observation for potential cleanup.

src/stores/audio/music_player.rs
patrickulrich: The code is correct as-is. The "race condition" cannot manifest in single-threaded Dioxus reactive code. If you want belt-and-suspenders safety, you could add a comment, but it's not worth changing the pattern.

src/stores/audio/music_player.rs
patrickulrich: The prev_index != snapshot.current_index guard is intentional and correct. Here's why:
1. The function is a polling snapshot — it runs every 1000ms (mod.rs:461) to sync Android native player state back to the Rust store. On every call, current_index in the snapshot will almost always be the same as the stored value (user isn't skipping tracks every second).
2. Without the guard, every poll would unconditionally write to current_index and current_track, triggering reactive updates across the UI every second — for no reason when nothing changed.
3. The stale-track scenario the reviewer raised doesn't apply here. The fallback logic in try_next_stream_mobile() (line 1095-1096) mutates current_track.media_url directly in the store. It does not mutate the playlist entries. So the snapshot sync doesn't need to re-read the playlist to pick up URL changes — those are already in current_track.
4. Playlist entries themselves are never mutated in-place. The playlist is only modified via extend() (append) or wholesale replacement in play_track(). Individual entries in the playlist vec don't get their media_url updated.
Verdict: The guard is intentional and correct. It prevents unnecessary reactive churn on a 1-second polling loop. No fix needed.

src/stores/media/blossom_store.rs
patrickulrich: The PR review missed the http_client() implementation.
All three functions (fetch_server_list, delete_from_server, mirror_to_server) now use the shared http_client() from src/platform/http.rs, which:
- Native platforms: sets a 15-second timeout (line 18) — actually longer than the old 10-second one
- WASM/browser: relies on the browser's built-in fetch timeout (the reqwest timeout API isn't available on wasm32)
The old per-request AbortController + setTimeout JavaScript pattern was replaced by a consolidated Rust-level timeout that's applied to every request through the shared client. No fix needed.

src/utils/parsing/thread_tree.rs
patrickulrich: The change was intentional and aligns with NIP-10:
1. marker=Root means "parent is the thread root" — the new code correctly treats it as the parent. The old code fell through to the last e tag, which could be a mention and not the actual parent.
2. The reviewer's example [e(root), e(mention), e(mention)] — the old behavior (returning a mention as parent) was the bug. These notes are replies to the root, and should be attached as direct children of the root in the thread tree.
3. Positional (unmarked) tags still follow the same "last e tag" fallback as before.
No fix needed. This was a genuine improvement in NIP-10 compliance.

Issue #313, opened 2026-05-08
Bug: Music playlist page shows "Invalid playlist address" for valid naddr links


goodmorningbitcoin:
## Description

When clicking a music playlist link (e.g., from a rendered `naddr` in a note or from `/notifications`), the `/music/playlist/:naddr` page shows **"Invalid playlist address"** instead of loading the playlist. The playlist event exists and is valid.

## Reproduction

1. View a note containing a `nostr:naddr1...` link for a kind 34139 playlist
2. The content parser renders it as a clickable playlist card
3. Click it → navigates to `/music/playlist/naddr1qvzqq...`
4. Page shows "Invalid playlist address" error

**Example URL:** `nostr.blue/music/playlist/naddr1qvzqqqy9tvpzp8rnydh4vh4ep02xs8pfawednsdcpe8psnxmlpxkmf6zaxgxwr9kqy88wumn8ghj7mn0wvhxcmmv9uq3wamnwvaz7tmjv4kxz7fwwpexjmtpdshxuet59uqjqamnwvaz7tmwdaehgu3wwdshg6r0v9exgetj9e3k7mf0d9mx7uneqqd8x6rpv3hhwcnpdehx2epdxymnwdeexversdfcxycnyvrvqax`

This decodes to:
- **Kind:** 34139 (playlist)
- **Pubkey:** `9c73236f565eb90bd4681c29ebb2d9c1b80e4e184cdbf84d6da742e990670cb6`
- **d-tag:** `shadowbanned-1777932858112`
- **Relays:** `wss://nos.lol/`, `wss://relay.primal.net/`, `wss://nostr.sathoarder.com/ivory`

The event exists and is fetchable — confirmed via direct relay query.

## Root Cause

`MusicPlaylistDetail` in `src/routes/music/playlist_detail.rs` (lines 17-24) expects the `naddr` parameter in `kind:pubkey:d-tag` colon-separated format:

```rust
let parts: Vec<&str> = naddr_clone.split(':').collect();
if parts.len() < 3 {
    error_msg.set(Some("Invalid playlist address".to_string()));
    loading.set(false);
    return;
}
let pubkey = parts[1];
let d_tag = parts[2..].join(":");
```

But the route receives the **bech32-encoded naddr** string (e.g., `naddr1qvzqq...`) from two callers:

[...]

Issue #312, opened 2026-05-08
Bug: Parent note not shown when navigating to reply from /notifications


goodmorningbitcoin:
## Description

When clicking a reply notification from `/notifications`, the `/note/:id` page opens but does not display the parent note above the reply. The expected behavior is that the parent (the note being replied to) renders above the opened note.

## Reproduction

1. Receive a reply notification
2. Click the reply from `/notifications`
3. Navigate to `/note/:hexid` — the reply loads but the parent note is missing above it

**Example note:** `74a75799a2139b5bf99c39242e635f2f839ba31388ba7ffe488cc009fd2bcd97`

This reply has a single `e` tag pointing to the root:
```json
["e", "203d8f4c4fae919796fb6dc73ebf159927e294b0087a32020f5ae6fe93fd0650", "wss://relay.primal.net/", "root", "0f563fe2cfdf180cb104586b95873379a0c1fdcfbc301a80c8255f33d15f039d"]
```

The parent event exists on `wss://relay.primal.net/` and is fetchable, but never loads.

## Root Cause

`fetch_parents_with_hints()` in `src/routes/note.rs` calls `relay::connection::fetch_events_from_relays()` with relay hint URLs extracted from `e` tags, **but does not first connect those relays to the pool**.

The nostr-sdk's `client.fetch_events_from()` only queries relays that are already in the connection pool (`internal_relay()` returns `Error::RelayNotFound` if the URL isn't present). When the hinted relay (e.g., `wss://relay.primal.net/`) is not one of the user's configured/pool relays, the fetch silently fails and the parent is never retrieved.

The error is swallowed at `note.rs:132-141`:
```rust
if let Ok(events) = relay::connection::fetch_events_from_relays(
    &nostr_client::get_client().unwrap(),
    filter,
    vec![relay_url.clone()],
    Duration::from_secs(5),
)
.await
{
    new_events.extend(events); // never reached if relay not in pool
}
```

If the hinted fetch fails, there is **no fallback** — the parent IDs are only in `hinted_grouped` and never added to `unhinted_ids` for the outbox/author-relay fallback path.

## Existing Correct Pattern

The codebase already handles this correctly in other places using `connect_ephemeral_relays`:

- **`fetch_event_targeted()`** in `src/stores/nostr_client/fetching.rs` (lines 399-408, 420-430) — connects relays ephemerally before fetching, then cleans up
[...]

Issue #311, opened 2026-05-07
Blocked users appear in conversation threads — ThreadedComment lacks block/mute filtering


goodmorningbitcoin:
## Bug

Blocked users' replies are fully visible in conversation threads. The main note correctly shows "Post from blocked user," but all replies in the thread render with full content, avatar, and display name regardless of block status.

## Root Cause

`src/components/threaded_comment.rs` has **zero block/mute awareness**:

- No `cached_muted_posts` or `cached_blocked_users` props
- No `is_blocked` / `is_muted` signals
- No "hidden" state or "Show anyway" toggle
- No call to `is_user_blocked_cached()` or `is_post_muted_cached()`

Meanwhile, `src/components/note_card.rs` (lines 409-553) has full block/mute handling: it accepts the caches as props, checks them, and renders a hidden state with "Post from blocked user" / "Muted post" and a "Show anyway" button.

The gap: `ThreadedComment` is used to render **every reply** in conversation threads, but it was never wired into the block system.

## Affected Pages (7+ routes)

`ThreadedComment` is used in all of these, meaning blocked users appear in reply threads on:

| Route | File |
|-------|------|
| Note detail | `src/routes/note.rs` |
| Article detail | `src/routes/articles/article_detail.rs` |
| NIP detail | `src/routes/nips/nip_detail.rs` |
| Photo detail | `src/routes/photos/photo_detail.rs` |
| Poll view | `src/routes/polls/poll_view.rs` |
| Video detail | `src/routes/video/video_detail.rs` |
| Voice message detail | `src/routes/voice/voice_message_detail.rs` |

### Additional Gap: Topics Subsystem

These files also lack block/mute filtering:
- `src/routes/topics/post_detail.rs` — topic post replies
- `src/components/topic/thread_view.rs` — topic thread rendering
- `src/components/topic/post_card.rs` — topic post cards
- `src/routes/topics/home.rs`, `browse.rs`, `topic_feed.rs` — topic feeds

And live chat components:
- `src/components/live/chat.rs` — live stream chat
- `src/components/live/channel_chat.rs` — channel chat

### 1. Add block/mute support to `ThreadedComment` (primary fix)

In `src/components/threaded_comment.rs`:

**a) Add props:**

```rust
pub fn ThreadedComment(
    node: ThreadNode,
    depth: usize,
    on_reply: Option<EventHandler<NostrEvent>>,
    precomputed_counts: Option<InteractionCounts>,
[...]

Issue #305, opened 2026-04-30
Bug: Pages load scrolled halfway down on fresh navigation


goodmorningbitcoin:
## Description

When navigating to pages (especially Home) via sidebar links or regular clicks, the page loads scrolled halfway down instead of starting at the top. This should only happen when using browser-back navigation.

## Root Cause

The `HOME_SCROLL_ANCHOR` scroll restore logic in `src/routes/home/mod.rs` (lines ~920-940) cannot distinguish between a **popstate** navigation (browser back) and a **push** navigation (link click).

### Broken flow:

1. User scrolls Home to 500px
2. `last_scroll_y` signal tracks this via a 200ms poll loop
3. User clicks a note → `navigator.push()` → Dioxus scrolls to top, then `use_drop` fires and saves `scroll_y=500, is_set=true` into `HOME_SCROLL_ANCHOR`
4. Later, user clicks "Home" in sidebar → fresh push, not browser back
5. Dioxus scrolls to 0 ✅
6. Home content loads → `use_effect` fires, sees `is_set == true` and matching `feed_type_label` → **restores scroll to 500px** 🐛

The code comment at line 896 explicitly acknowledges the timing issue — the signal intentionally captures the pre-navigation scroll value so `use_drop` always has a non-zero position to save. But this means the anchor is always set on unmount, regardless of navigation type.

## Suggested Fix

Add a `popstate` detection flag so the restore logic only fires on browser-back:

### 1. `src/stores/ui/scroll_restore.rs` — add popstate detection

```rust
/// Set up popstate detection flag (call once at app startup)
pub fn setup_popstate_flag() {
    let _ = document::eval(
        r#"window.__nostrBlueWasPopstate = false;
           window.addEventListener("popstate", () => {
               window.__nostrBlueWasPopstate = true;
           });"#
    );
}

/// Check and consume the popstate flag
pub async fn was_popstate_nav() -> bool {
    let result = document::eval("return window.__nostrBlueWasPopstate === true")
        .await
        .ok()
        .and_then(|v| v.as_bool())
        .unwrap_or(false);
    if result {
        let _ = document::eval("window.__nostrBlueWasPopstate = false");
    }
    result
}
```

### 2. `src/routes/home/mod.rs` — guard the restore effect

```rust
// Restore scroll position after feed content loads (only for popstate/back)
use_effect(move || {
    if let DataState::Loaded(items) = &*feed_state.read() {
        if !items.is_empty() {
[...]

PR #314, opened 2026-05-09
Release 0.8.11: add NIP-5A nsites, Blobbi pet system, music content cards, and thread moderation

patrickulrich:
- **NIP-5A Static Pages (nsites)** — Add nsite hosting support with repo pages panel, directory hub, manifest publishing, and 7 gateway support to the /code module
- **Blobbi Virtual Pet Overhaul** — Companion behavior loop with gaze tracking, composable visual recipe system, mission tracking, hatching ceremony, and expanded care actions
- **Music Content Cards** — Shared nostr.blue music URLs (albums, tracks, artists) now render as interactive cards with play buttons instead of plain text links

## Fixes
- Add block/mute filtering to conversation threads, topic components, and live chat with "Show anyway" toggle (#311)
- Connect ephemeral relays before targeted parent/reply fetches to fix silent failures on non-pool relays (#312)
- Fix scroll-to-top on push navigation, restore scroll only on browser back (#305)
- Centralize kind-to-route mapping; fix broken share links for verts, wikis, pinboards, and livestreams
- Fix naddr parsing for playlists and consolidate 6 duplicated parse functions (#313)
- Allow hyphens in NIP-30 custom emoji shortcodes per spec
- Skip signing kind 10011 event when NIP-39 identities are unchanged
- Show "Sign In Required" page for unauthenticated Podcast Index API pages
- Deduplicate parent events in fetch_parents_with_hints to fix Dioxus keyed sibling panic


src/components/topic/post_card.rs
```diff
@@ -24,7 +27,58 @@ pub fn TopicPostCard(
...
+        author_pubkey_for_check,
+    )| {
+        let post_id = post_id_for_check.clone();
+        let author_pubkey = author_pubkey_for_check.clone();
+        if let Some(ref muted_set) = cached_muted_posts_reactive {
+            if let Ok(muted) = nostr_client::is_post_muted_cached(&post_id, muted_set) {
+                is_muted.set(Some(muted));
+            }
+        }
+        if let Some(ref blocked_set) = cached_blocked_users {
```
goodmorningbitcoin: 🐛 **Bug: Wrong variable — broken reactive subscription**

This uses `cached_blocked_users` (the original prop) instead of `cached_blocked_users_reactive` (the reactive clone). The `use_reactive!` macro captures `cached_blocked_users_reactive` as its reactive dependency, but the actual check on this line reads from the non-reactive `cached_blocked_users`.

This means:
1. The blocked-user cache check inside `use_reactive` reads a **stale** value — the one captured before the effect was set up
2. When the cache updates, `use_reactive` re-fires (because `cached_blocked_users_reactive` changed), but this line still reads the old prop value

Compare with `threaded_comment.rs` which correctly uses `cached_blocked_users_reactive` here.

**Fix:** Change `cached_blocked_users` → `cached_blocked_users_reactive` on this line.

src/components/live/channel_chat.rs
```diff
@@ -426,6 +426,37 @@ pub fn ChannelChat(channel_id: String) -> Element {
 #[component]
 fn ChannelChatMessage(event: Event) -> Element {
     let author_pubkey = event.pubkey.to_string();
+    let (_cached_muted_posts, cached_blocked_users) = use_mute_block_cache();
```
goodmorningbitcoin: ⚠️ **Performance: Per-message hook instance**

`use_mute_block_cache()` is called inside `ChannelChatMessage`, which means each message component creates its own `use_signal` instances and `use_effect`. In Dioxus, hooks are per-component-instance — so 100 visible messages = 100 independent signal sets + 100 independent fetch effects.

The threaded components (`ThreadedComment`, `ThreadView`) do this correctly: the parent calls the hook once and passes `Rc<HashSet<String>>` down as props. Live chat should follow the same pattern.

**Recommendation:** Call `use_mute_block_cache()` once in the parent `ChannelChat` component and pass the signals (or the `.read().clone()` values) as props to each `ChannelChatMessage`.

src/components/live/chat.rs
```diff
@@ -334,6 +334,37 @@ pub fn LiveChat(stream_author_pubkey: String, stream_d_tag: String) -> Element {
 #[component]
 fn ChatMessage(event: Event) -> Element {
     let author_pubkey = event.pubkey.to_string();
+    let (_cached_muted_posts, cached_blocked_users) = use_mute_block_cache();
```
goodmorningbitcoin: ⚠️ **Same as `channel_chat.rs` above** — `use_mute_block_cache()` is called per `ChatMessage` instance instead of once at the parent level. Consider hoisting to the parent `Chat` component.

src/routes/code/code_repo_pages.rs
```diff
@@ -0,0 +1,60 @@
...
+                                "Static Pages: {name}"
+                            }
+                            p { class: "text-muted-foreground mt-1",
+                                "Publish and manage static pages for this repository."
+                            }
+                        }
+                        RepoPagesPanel {
+                            repo: r.clone(),
+                            naddr: naddr.clone(),
+                            is_owner: true,
```
goodmorningbitcoin: ⚠️ **Hardcoded `is_owner: true`**

The `RepoPagesPanel` receives `is_owner: true` unconditionally. This means any user viewing any repo's Pages tab gets the full owner experience (site name input, push manifest button, etc.).

The button is gated by `HAS_SIGNER`, so non-authed users can't actually push. But the cosmetic issue remains — non-owners see UI elements they shouldn't.

**Fix:** Compare the repo's pubkey against the logged-in user's pubkey (same pattern as the code repo overview tab) to determine actual ownership.

src/components/live/channel_chat.rs
```diff
@@ -426,6 +426,37 @@ pub fn ChannelChat(channel_id: String) -> Element {
...
+                    is_author_blocked.set(Some(blocked));
+                }
+            }
+            if cached_blocked_users_val.is_none() {
+                spawn(async move {
+                    match nostr_client::is_user_blocked(author_pubkey).await {
+                        Ok(blocked) => is_author_blocked.set(Some(blocked)),
+                        Err(_) => is_author_blocked.set(Some(false)),
+                    }
+                });
```
goodmorningbitcoin: 💡 **UX inconsistency: No "Show anyway" for blocked chat messages**

Threaded components (`ThreadedComment`, `TopicPostCard`) show a "[Blocked user] — Show anyway" toggle. Live chat silently returns `rsx! {}` — the message just disappears with no indication it existed.

For consistency and transparency, consider showing a collapsed "Message from blocked user" indicator with a toggle, matching the threaded pattern.

goodmorningbitcoin (commented):
## Code Review — PR #314

### Summary

Solid release PR. The core infrastructure fixes are well-implemented:
- **Ephemeral relay pattern** (#312) correctly pairs connect/cleanup across all 4 call sites in `note.rs`
- **Shared `parse_naddr`** (#313) properly consolidates 6 duplicate implementations with bech32 + legacy fallback
- **Block/mute filtering** (#311) consistent across threads, topics, and live chat with "Show anyway" toggle
- **NIP-5A nsites** clean implementation with correct kind numbers, slug validation, reserved subdomain protection
- **Scroll-to-top fix** clever popstate-based approach
- **Kind-to-route consolidation** excellent deduplication

### Issues Found

| Severity | File | Issue |
|---|---|---|
| 🐛 **Bug** | `post_card.rs:54` | `cached_blocked_users` used instead of `cached_blocked_users_reactive` — broken reactive subscription for blocked-user check |
| ⚠️ **Perf** | `channel_chat.rs:429`, `chat.rs:337` | `use_mute_block_cache()` called per message instance, creating N independent fetch effects instead of one shared cache at parent level |
| ⚠️ **Correctness** | `code_repo_pages.rs:53` | `is_owner: true` hardcoded — any user viewing a repo's Pages tab gets the owner panel including site name input |
| 💡 **Minor** | `channel_chat.rs:451`, `chat.rs:362` | Blocked messages return empty `rsx!{}` — no "Show anyway" toggle unlike the threaded components |
| 💡 **Minor** | `route_for_kind.rs` | Kind 30311 (livestream) now routes naddrs via `LiveStreamDetail { note_id: naddr }` — verify the route handler can parse naddr-formatted values |

See inline comments for details.

**Verdict: Approve with 1 bug fix needed** (the `cached_blocked_users` reactive prop issue in `post_card.rs`).

src/components/topic/post_card.rs
patrickulrich: Valid.

src/components/live/channel_chat.rs
patrickulrich: Valid.

src/components/live/chat.rs
patrickulrich: Valid.

src/routes/code/code_repo_pages.rs
patrickulrich: Valid.

src/components/live/channel_chat.rs
patrickulrich: Valid.

src/utils/route_for_kind.rs
```diff
@@ -0,0 +1,121 @@
...
+        30617 => Some(Route::CodeRepo { naddr }),
+        30818 => {
+            let npub = pubkey.to_bech32().unwrap_or_else(|_| pubkey.to_hex());
+            Some(Route::WikiDetail { npub, identifier })
+        }
+        30030..=30033 => Some(Route::CitationDetail { naddr }),
+        31922 | 31923 => Some(Route::CalendarEventDetail { naddr, from: None }),
+        32123 => Some(Route::MusicPlaylistDetail { naddr }),
+        33889 => Some(Route::PinBoardDetail { naddr }),
+        34235 | 34236 => Some(Route::VideoDetail { video_id: naddr }),
```
goodmorningbitcoin, line 30: 🐛 **Bug: Missing kind 30067 for Pinboard routing**

Pinboards use kind `30067` throughout the codebase (`KIND_PINBOARD = 30067` in `pin_boards_store/mod.rs`, `mentions.rs:544`). The old `mentions.rs` correctly mapped `30067 → PinBoardDetail`. This centralized function maps `33889 → PinBoardDetail` but has **no entry for 30067**.

**Impact:** Pinboard naddr links in mentions render as plain text. Share URLs for pinboards break.

**Fix:** Add `30067 => Some(Route::PinBoardDetail { naddr })` to this match, and `30067 => "Pinboard"` to `content_label_for_kind()`.

src/utils/nips/nip5a.rs
```diff
@@ -0,0 +1,472 @@
...
+    ".nuxt",
+    ".output",
+    ".vercel",
+    ".netlify",
+];
+
+const STATIC_EXTENSIONS: &[&str] = &[
+    "html", "htm", "css", "js", "mjs", "cjs", "json", "webmanifest", "map", "txt", "md", "xml",
+    "svg", "png", "jpg", "jpeg", "gif", "webp", "ico", "webp", "woff", "woff2", "ttf", "otf",
+    "wasm", "webm", "mp4", "weba", "mp3", "ogg", "pdf",
```
goodmorningbitcoin, line 336: 🧹 **Duplicate `"webp"` in STATIC_EXTENSIONS**

`"webp"` appears twice in the array (positions 7 and 9). Harmless but messy. Remove the duplicate.

src/utils/nips/nip5a.rs
```diff
@@ -0,0 +1,472 @@
...
+            }
+        } else if kind == TagKind::Custom(Cow::Borrowed("source")) {
+            if let Some(s) = tag.content() {
+                source = Some(s.to_string());
+            }
+        }
+    }
+
+    if let Some(t) = &title {
+        if title.is_none() {
```
goodmorningbitcoin: 🧹 **Dead code in `parse_manifest`**

```rust
if let Some(t) = &title {
    if title.is_none() {  // ← always false when outer is Some
        title = Some(t.clone());
    }
}
```

If `title` is `Some`, the inner `title.is_none()` is always `false` — this block never executes. Leftover from a merge. Remove it.

src/components/blobbi/core/decay.rs
```diff
@@ -166,13 +207,13 @@ pub fn should_emit_shell_penalty(pet: &BlobbiCompanion) -> bool {
...
+#[allow(dead_code)]
 pub fn shell_penalty_hours(pet: &BlobbiCompanion, now_secs: u64) -> f64 {
     let shell = pet.shell_integrity.unwrap_or(STAT_MAX);
     if shell >= 50.0 {
         return 0.0;
     }
-    let rate = 2.0 + 2.0 * ((50.0 - shell) / 50.0).min(1.0);
-    let _hours_to_drop = shell / rate;
+    let _rate = 2.0 + 2.0 * ((50.0 - shell) / 50.0).min(1.0);
     let elapsed = now_secs.saturating_sub(pet.last_decay_at.unwrap_or(now_secs)) as f64 / 3600.0;
```
goodmorningbitcoin: ⚠️ **`shell_penalty_hours` computes rate but never uses it**

`_rate` is computed but the function returns raw `elapsed` (time since last decay). The rate calculation has zero effect on the return value.

Either this should return penalty hours based on the rate, or the dead code should be removed.

src/components/rich_content/nostr_blue_renderers.rs
```diff
@@ -1061,3 +1063,308 @@ pub(super) fn NostrBlueChannelRenderer(id: String) -> Element {
...
+                        {
+                            render_music_compact_card(
+                                image_clone.as_deref(),
+                                &title_clone,
+                                &format!("{} · {} {}", artist_clone, count_val, if count_val == 1 { "track" } else { "tracks" }),
+                                move |_: MouseEvent| {
+                                    let fid = fid;
+                                    spawn(async move {
+                                        let feed = match podcast_index::get_podcast_by_id(fid).await {
+                                            Ok(f) => f,
```
goodmorningbitcoin: ⚠️ **Double-fetch on play button click**

The `use_resource` above already fetches `get_podcast_by_id` and `get_episodes_by_feed_id` for display. The play button `onclick` handler fetches **both again** inside this `spawn`. The display data is already available in the `resource` signal.

**Fix:** Store the fetched episodes in a signal or use the already-fetched data from the resource instead of re-fetching.

goodmorningbitcoin:
## Additional issues found (pre-existing code, not in this PR's diff)

These issues exist in code that wasn't changed by this PR, but are worth noting for a follow-up:

---

### 🐛 Bug: `evolve_baby` visual effects never trigger (`stage_transition.rs:239`)

```rust
if let Some(eff) = generate_random_visual_effect(0.5) {
```

`generate_random_visual_effect()` only returns `Some` when `roll <= 0.15` (see `seed.rs:199`). Passing `0.5` means this is dead code — evolving Blobbis **never** get a random visual effect assigned.

**Fix:** `generate_random_visual_effect(rand::random::<f64>() * 0.15)` or any value <= 0.15.

---

### ⚠️ Field/method name collision: `is_sleeping` (`types.rs:154` and `types.rs:211`)

The struct has a **field** `is_sleeping: bool` (line 154) AND a **method** `is_sleeping(&self) -> bool` (line 211). The method checks both the field AND `self.state == BlobbiState::Sleeping`.

Inside `impl` methods, `self.is_sleeping` accesses the field directly. Outside, `blobbi.is_sleeping()` invokes the method. These can silently diverge — any code that sets `state = Sleeping` without the flag gets correct results from the method but wrong results from the field.

**Recommendation:** Rename the field to `sleeping_flag` or derive it solely from `BlobbiState`.

---

### ⚠️ No `add_blobbi_to_collection` function (`blobbi_store.rs`)

The store has `set_collection`, `update_blobbi_in_collection`, but no `add_blobbi_to_collection`. New pets (hatching, breeding) have no way to be added without replacing the entire collection. `update_blobbi_in_collection` only updates existing entries by matching `d` tag.

**Fix:** Add an `add_blobbi_to_collection(blobbi: BlobbiCompanion)` function that pushes to the collection signal.

---

### 💡 Missing kind 30030-30033 in `content_label_for_kind` (`route_for_kind.rs:95-120`)

`route_for_naddr` maps kinds 30030-30033 to `CitationDetail`, but `content_label_for_kind` has no entry for them. They show as generic "Content" in share modals and notification labels.

**Fix:** Add `30030..=30033 => "Citation"` to the match arm.

PR #322, opened 2026-05-11
Release 0.8.12: Relay Feeds, Performance Overhaul, Video Thumbnails & Fixes

### LINKED ISSUE #318: Detail pages fail with "too many relays" — naddr relay hints discarded, gossip hits pool cap
## Bug Description

Opening any addressable event detail page (articles, livestreams, recipes, badges, pinboards, calendar events, code repos, podcasts, publications) fails with:

```
Error Loading Article
Failed to fetch event: too many relays
```

The relay connection pool has an artificial 50-relay cap. When gossip mode discovers new relays, it permanently adds them to the pool until it hits the cap and throws `Error::TooManyRelays`.

### Bug 1: Artificial relay pool cap should not exist

`stores/nostr_client/mod.rs:184` sets:

```rust
.pool(RelayPoolOptions::new().max_relays(Some(50)));
```

The SDK defaults to `None` (unlimited). This cap only exists as a bandage for gossip leaking permanent connections into the pool (Bug 3). If relay connections are properly cleaned up after use, the pool only holds the user's intentionally configured relays and never grows unboundedly. **Remove the limit entirely.**

### Bug 2: Relay hints from naddr are discarded

Every naddr bech32 can contain relay hints — relay URLs where the event is known to be available. Currently, **every detail page except Note Detail** decodes the naddr to extract `(pubkey, identifier)` and **silently drops the relay hints**. Without hints, the fetch can't target specific relays and falls through to a gossip-based broadcast that bloats the pool.

### Bug 3: Gossip mode permanently adds relays with no cleanup

The rust-nostr SDK stores all relays in a **single flat HashMap** — no separation between user-configured relays and dynamically discovered ones. When `client.fetch_events()` is called with gossip enabled, `break_down_filter()` in `nostr-sdk/src/client/mod.rs` permanently adds every discovered relay via `add_gossip_relay()`:

```rust
// From nostr-sdk internals:
for url in filters.keys() {
    if self.add_gossip_relay(url).await? {  // permanently added
        self.connect_relay(url).await?;
    }
}
```

These gossip relays get the `GOSSIP` flag which **protects them from removal** via normal `remove_relay()`. Only `force_remove_relay()` can clean them up. Over a session, browsing profiles and feeds accumulates relay connections with no automatic cleanup.

## Call Flow (article example)

```
ArticleDetail(naddr)
  -> decode_naddr() -> (pubkey, identifier)           // relay hints DROPPED
  -> fetch_event_by_coordinate(kind, pubkey, id)      // no relay hints passed
  -> fetch_event_by_coordinate_with_relays(..., [])   // empty relay_hints
  -> client.fetch_events(filter, timeout)              // gossip path
  -> break_down_filter() -> add_gossip_relay()          // permanently added to pool
  -> Error::TooManyRelays { limit: 50 }                // BOOM
```

## Affected Pages

All detail pages that decode naddr, drop hints, then call `fetch_events_aggregated` or `client.fetch_events`:

| Page | File | Decoder | Dropped at |
|------|------|---------|-----------|
| Article Detail | `routes/articles/article_detail.rs` | own `decode_naddr` | L452 — never extracts `.relays` |
| Livestream Detail | `routes/video/live_stream_detail.rs` | own `parse_naddr` | L371 — returns `(pubkey, id)` |
| Recipe Detail | `stores/content/recipe_store.rs` | `Coordinate::from_bech32` | L341 — no relay field |
| Badge Detail | `utils/nips/nip58.rs` | `parse_badge_naddr` | L215 -> `fetch_events_aggregated` |
| Pinboard Detail | `stores/social/pin_boards_store/fetch.rs` | `Coordinate::from_bech32` | L237 — no relay field |
| Calendar Event | `stores/calendar_store/fetch.rs` | `Nip19Coordinate::from_bech32` | L580-584 — ignores `.relays` |
| Code Repository | `services/git_hosting/repository.rs` | `decode_repo_naddr` | L18 — `_relay_hints` unused |
| Podcast Episode | `routes/podcast/podcast_episode_detail.rs` | custom | `fetch_events_aggregated` |
| Podcast Nostr | `routes/podcast/podcast_nostr_detail.rs` | custom | `fetch_events_aggregated` |
| Publication Detail | `routes/articles/publication_detail.rs` | custom | `fetch_events_aggregated` |

## Not Affected

- **Note Detail** (`routes/note.rs`) — uses `fetch_event_targeted` with proper 5-phase ephemeral relay pattern
- **Embedded cards** in `nostr_blue_renderers.rs` — use the safe `use_fetch_event_by_coordinate_with_message` hook which properly extracts and passes relay hints

### Step 1: Remove the relay pool limit

Remove `max_relays(Some(50))` from both client initialization paths in `stores/nostr_client/mod.rs` (lines 184 and 235). Let the SDK default to unlimited. With proper ephemeral cleanup, the pool only holds user-configured relays.

### Step 2: Rewrite `fetch_event_by_coordinate_with_relays` to use ephemeral pattern

Rewrite the function at `stores/relay/connection.rs:277` to follow the same 5-phase approach as `fetch_event_targeted`:

1. **DB cache check** (already done)
2. **Relay hints** — use `fetch_events_from(hint_urls, filter, timeout)` to query specific relays directly, bypassing gossip entirely. Then `force_remove_relay()` any that were newly added.
3. **Broadcast to connected relays** — query relays already in the pool via `fetch_events_from(pool_urls, filter, timeout)`, no new connections needed
4. **Author relays** — resolve NIP-65 write relays, ephemeral fetch via `fetch_events_from`, then `force_remove_relay()` cleanup
5. **Gossip fallback** — `client.fetch_events()` as last resort only

The key principle: **never use `client.fetch_events()` with gossip for targeted single-event lookups**. Use `client.fetch_events_from(specific_urls, filter, timeout)` instead, which adds nothing to the pool. This is the SDK's own recommended pattern from their `aggregated-query.rs` example:

```rust
// Add temp relay and fetch
client.add_relay("wss://example.com").await?;
client.connect_relay("wss://example.com").await?;
let events = client
    .fetch_events_from(["wss://example.com"], filter, Duration::from_secs(10))
    .await?;
client.force_remove_relay("wss://example.com").await?;
```

### Step 3: Wire relay hints from naddr to the fetch function

Every detail page that decodes an naddr should extract `coord.relays` and pass them through to `fetch_event_by_coordinate_with_relays`. Options:
- **A)** Migrate pages to use the shared `use_fetch_event_by_coordinate_with_message` hook (already correct)
- **B)** Create a shared `fetch_event_by_naddr(naddr)` function that decodes + extracts hints + calls the relay-aware fetch

### Step 4: Periodic gossip relay cleanup

After any operation that does go through gossip (feed loading, etc.), clean up accumulated gossip relays:

```rust
let all = client.pool().all_relays().await;
for (url, relay) in all {
    let flags = relay.flags();
    // Only remove relays that have GOSSIP flag but no user-configured flags
    if flags.has_gossip() && !flags.has_any(READ | WRITE | DISCOVERY) {
        client.force_remove_relay(&url).await?;
    }
}
```

This keeps the pool clean while preserving user-configured relays.

## Architecture Principle

Relay connections should be treated like a cache, not a landfill:

- **User-configured relays** — permanent, auto-reconnect, the backbone of the pool
- **Discovered relays** — ephemeral, used for one fetch, cleaned up immediately after
- **Relay hints from naddr/nevent** — the primary source for targeted lookups, never discarded
- **Gossip discovery** — fallback only, never the first path, and always cleaned up after

## Reproduction

1. Use nostr.blue for a session, browsing feeds and profiles until the relay pool fills up
2. Navigate to any article (e.g., one shared as an naddr link)
3. Observe "Error Loading Article — Failed to fetch event: too many relays"

The article IS available on its relay hints — the fetch just never tries them.

patrickulrich:
- **Relay Feeds** — Per-relay content browsing with real-time subscriptions
- **Performance** — Faster interaction loading (batched, DB-fast-path, streaming) and instant reply loading (DB-first with parallel fetching)
- **Video Thumbnails** — Cross-platform blurhash + canvas capture with media fragment preloading
- **Relay Connection Fix** — Fix detail pages failing with "too many relays" error (closes #318)
- **Post-Login Init** — Await user relays before NIP-78 loads, parallelize initialization
- **AI Chat** — Fix tool use content overflowing message bubbles
- **Pinboard Routing** — Fix kind 30067 routing, remove dead code, eliminate podcast double-fetch
- **Reactivity Fixes** — Fix reactive bugs, per-message hook overhead, hardcoded is_owner, and blocked-user UX
- **Blobbi** — Remove unused visual_effect feature, fix is_sleeping serialization
- **CI** — Populate docs submodules and resolve relative URLs for reqwest WASM

### DIFF
--- a/src/components/article_content.rs
+++ b/src/components/article_content.rs
@@ -1,52 +1,91 @@
+use crate::components::rich_content::NostrUriRenderer;
 use crate::components::SensitiveContent;
-use crate::utils::markdown::render_markdown;
+use crate::utils::markdown::{extract_nostr_uris, render_markdown};
 use dioxus::prelude::*;
+
+const PROSE_CLASSES: &str = "article-content prose prose-lg prose-neutral dark:prose-invert max-w-none break-words
+    [&_h1]:text-4xl [&_h1]:font-bold [&_h1]:mt-8 [&_h1]:mb-4
+    [&_h2]:text-3xl [&_h2]:font-bold [&_h2]:mt-6 [&_h2]:mb-3
+    [&_h3]:text-2xl [&_h3]:font-semibold [&_h3]:mt-5 [&_h3]:mb-2
+    [&_p]:my-4 [&_p]:leading-relaxed
+    [&_a]:text-primary [&_a]:underline hover:[&_a]:text-primary/80
+    [&_ul]:my-4 [&_ul]:pl-6 [&_ul]:list-disc
+    [&_ol]:my-4 [&_ol]:pl-6 [&_ol]:list-decimal
+    [&_li]:my-2
+    [&_blockquote]:border-l-4 [&_blockquote]:border-primary [&_blockquote]:pl-4 [&_blockquote]:my-4 [&_blockquote]:italic
+    [&_code]:bg-muted [&_code]:px-1 [&_code]:py-0.5 [&_code]:rounded [&_code]:text-sm
+    [&_pre]:bg-muted [&_pre]:p-4 [&_pre]:rounded-lg [&_pre]:overflow-x-auto [&_pre]:my-4
+    [&_img]:max-w-full [&_img]:h-auto [&_img]:rounded-lg [&_img]:my-6
+    [&_table]:w-full [&_table]:my-4
+    [&_th]:border [&_th]:border-border [&_th]:bg-muted [&_th]:px-4 [&_th]:py-2 [&_th]:font-semibold
+    [&_td]:border [&_td]:border-border [&_td]:px-4 [&_td]:py-2";
+
+enum ContentSegment {
+    Html(String),
+    NostrUri(usize),
+}
+
+fn split_html_on_markers(html: &str, uri_count: usize) -> Vec<ContentSegment> {
+    let mut segments = Vec::new();
+    let mut remaining = html;
+    for idx in 0..uri_count {
+        let marker = format!("%%NOSTR_BLUE_EMBED_{}%%", idx);
+        if let Some(pos) = remaining.find(&marker) {
+            if pos > 0 {
+                segments.push(ContentSegment::Html(remaining[..pos].to_string()));
+            }
+            segments.push(ContentSegment::NostrUri(idx));
+            remaining = &remaining[pos + marker.len()..];
+        }
+    }
+    if !remaining.is_empty() {
+        segments.push(ContentSegment::Html(remaining.to_string()));
+    }
+    segments
+}
+
 #[component]
-pub fn ArticleContent(content: String, #[props(default)] content_warning: Option<Option<String>>) -> Element {
-    let html_content = render_markdown(&content);
-    rsx! {
-        if let Some(reason) = content_warning {
-            SensitiveContent { reason,
-                div {
-                    dangerous_inner_html: "{html_content}",
-                    class: "article-content prose prose-lg prose-neutral dark:prose-invert max-w-none
-                           [&_h1]:text-4xl [&_h1]:font-bold [&_h1]:mt-8 [&_h1]:mb-4
-                           [&_h2]:text-3xl [&_h2]:font-bold [&_h2]:mt-6 [&_h2]:mb-3
-                           [&_h3]:text-2xl [&_h3]:font-semibold [&_h3]:mt-5 [&_h3]:mb-2
-                           [&_p]:my-4 [&_p]:leading-relaxed
-                           [&_a]:text-primary [&_a]:underline hover:[&_a]:text-primary/80
-                           [&_ul]:my-4 [&_ul]:pl-6 [&_ul]:list-disc
-                           [&_ol]:my-4 [&_ol]:pl-6 [&_ol]:list-decimal
-                           [&_li]:my-2
-                           [&_blockquote]:border-l-4 [&_blockquote]:border-primary [&_blockquote]:pl-4 [&_blockquote]:my-4 [&_blockquote]:italic
-                           [&_code]:bg-muted [&_code]:px-1 [&_code]:py-0.5 [&_code]:rounded [&_code]:text-sm
-                           [&_pre]:bg-muted [&_pre]:p-4 [&_pre]:rounded-lg [&_pre]:overflow-x-auto [&_pre]:my-4
-                           [&_img]:max-w-full [&_img]:h-auto [&_img]:rounded-lg [&_img]:my-6
-                           [&_table]:w-full [&_table]:my-4
-                           [&_th]:border [&_th]:border-border [&_th]:bg-muted [&_th]:px-4 [&_th]:py-2 [&_th]:font-semibold
-                           [&_td]:border [&_td]:border-border [&_td]:px-4 [&_td]:py-2",
+pub fn ArticleContent(
+    content: String,
+    #[props(default)] content_warning: Option<Option<String>>,
+) -> Element {
+    let (content_with_markers, nostr_uris) = extract_nostr_uris(&content);
+    let html_content = render_markdown(&content_with_markers);
+    let segments = split_html_on_markers(&html_content, nostr_uris.len());
+
+    let rendered = rsx! {
+        for (seg_idx, segment) in segments.into_iter().enumerate() {
+            match segment {
+                ContentSegment::Html(html) => {
+                    let key = format!("html-{seg_idx}");
+                    rsx! {
+                        div {
+                            key: "{key}",
+                            dangerous_inner_html: "{html}",
+                            class: PROSE_CLASSES,
+                        }
+                    }
+                }
+                ContentSegment::NostrUri(idx) => {
+                    let uri = nostr_uris[idx].clone();
+                    let key = format!("nostr-{idx}");
+                    rsx! {
+                        div {
+                            key: "{key}",
+                            class: "my-2",
+                            NostrUriRenderer { uri }
+                        }
+                    }
                 }
             }
+        }
+    };
+
+    rsx! {
+        if let Some(reason) = content_warning {
+            SensitiveContent { reason, {rendered} }
         } else {
-            div {
-                dangerous_inner_html: "{html_content}",
-                class: "article-content prose prose-lg prose-neutral dark:prose-invert max-w-none
-                       [&_h1]:text-4xl [&_h1]:font-bold [&_h1]:mt-8 [&_h1]:mb-4
-                       [&_h2]:text-3xl [&_h2]:font-bold [&_h2]:mt-6 [&_h2]:mb-3
-                       [&_h3]:text-2xl [&_h3]:font-semibold [&_h3]:mt-5 [&_h3]:mb-2
-                       [&_p]:my-4 [&_p]:leading-relaxed
-                       [&_a]:text-primary [&_a]:underline hover:[&_a]:text-primary/80
-                       [&_ul]:my-4 [&_ul]:pl-6 [&_ul]:list-disc
-                       [&_ol]:my-4 [&_ol]:pl-6 [&_ol]:list-decimal
-                       [&_li]:my-2
-                       [&_blockquote]:border-l-4 [&_blockquote]:border-primary [&_blockquote]:pl-4 [&_blockquote]:my-4 [&_blockquote]:italic
-                       [&_code]:bg-muted [&_code]:px-1 [&_code]:py-0.5 [&_code]:rounded [&_code]:text-sm
-                       [&_pre]:bg-muted [&_pre]:p-4 [&_pre]:rounded-lg [&_pre]:overflow-x-auto [&_pre]:my-4
-                       [&_img]:max-w-full [&_img]:h-auto [&_img]:rounded-lg [&_img]:my-6
-                       [&_table]:w-full [&_table]:my-4
-                       [&_th]:border [&_th]:border-border [&_th]:bg-muted [&_th]:px-4 [&_th]:py-2 [&_th]:font-semibold
-                       [&_td]:border [&_td]:border-border [&_td]:px-4 [&_td]:py-2",
-            }
+            {rendered}
         }
     }
 }
--- a/src/components/nip05_badge.rs
+++ b/src/components/nip05_badge.rs
@@ -1,0 +1,114 @@
+use crate::services::nip05::{self, Nip05Status};
+use dioxus::prelude::*;
+
+#[derive(Props, Clone, PartialEq)]
+pub struct Nip05BadgeProps {
+    pub pubkey: String,
+    pub nip05: String,
+}
+
+#[component]
+pub fn Nip05Badge(props: Nip05BadgeProps) -> Element {
+    let mut status = use_signal(|| Nip05Status::Unknown);
+    let pubkey_for_effect = props.pubkey.clone();
+    let nip05_for_effect = props.nip05.clone();
+
+    use_effect(use_reactive(
+        (&props.pubkey, &props.nip05),
+        move |(pubkey, nip05)| {
+            let current = nip05::get_nip05_status(&pubkey, &nip05);
+            status.set(current.clone());
+            if matches!(current, Nip05Status::Unknown) {
+                nip05::verify_nip05(&pubkey, &nip05);
+            }
+        },
+    ));
+
+    use_future(move || {
+        let pk = pubkey_for_effect.clone();
+        let nip = nip05_for_effect.clone();
+        async move {
+            loop {
+                crate::platform::timer::sleep_ms(200).await;
+                let current = nip05::get_nip05_status(&pk, &nip);
+                if *status.read() != current {
+                    status.set(current);
+                }
+            }
+        }
+    });
+
+    let s = status.read().clone();
+    match s {
+        Nip05Status::Verified => rsx! {
+            span {
+                class: "inline-flex items-center gap-1 text-green-500",
+                title: "NIP-05 verified",
+                svg {
+                    class: "w-4 h-4",
+                    xmlns: "http://www.w3.org/2000/svg",
+                    view_box: "0 0 24 24",
+                    fill: "none",
+                    stroke: "currentColor",
+                    stroke_width: "2",
+                    stroke_linecap: "round",
+                    stroke_linejoin: "round",
+                    path { d: "M22 11.08V12a10 10 0 1 1-5.93-9.14" }
+                    polyline { points: "22 4 12 14.01 9 11.01" }
+                }
+            }
+        },
+        Nip05Status::Impersonator => rsx! {
+            span {
+                class: "inline-flex items-center gap-1 text-red-500",
+                title: "NIP-05 verification failed: pubkey mismatch",
+                svg {
+                    class: "w-4 h-4",
+                    xmlns: "http://www.w3.org/2000/svg",
+                    view_box: "0 0 24 24",
+                    fill: "none",
+                    stroke: "currentColor",
+                    stroke_width: "2",
+                    stroke_linecap: "round",
+                    stroke_linejoin: "round",
+                    circle { cx: "12", cy: "12", r: "10" }
+                    line { x1: "15", y1: "9", x2: "9", y2: "15" }
+                    line { x1: "9", y1: "9", x2: "15", y2: "15" }
+                }
+            }
+        },
+        Nip05Status::Error => {
+            let pk = props.pubkey.clone();
+            let nip = props.nip05.clone();
+            rsx! {
+                span {
+                    class: "inline-flex items-center gap-1 text-orange-500 cursor-pointer",
+                    title: "NIP-05 verification failed. Click to retry.",
+                    onclick: move |_| {
+                        nip05::retry_nip05(&pk, &nip);
+                    },
+                    svg {
+                        class: "w-4 h-4",
+                        xmlns: "http://www.w3.org/2000/svg",
+                        view_box: "0 0 24 24",
+                        fill: "none",
+                        stroke: "currentColor",
+                        stroke_width: "2",
+                        stroke_linecap: "round",
+                        stroke_linejoin: "round",
+                        path { d: "M10.29 3.86L1.82 18a2 2 0 0 0 1.71 3h16.94a2 2 0 0 0 1.71-3L13.71 3.86a2 2 0 0 0-3.42 0z" }
+                        line { x1: "12", y1: "9", x2: "12", y2: "13" }
+                        line { x1: "12", y1: "17", x2: "12.01", y2: "17" }
+                    }
+                }
+            }
+        }
+        Nip05Status::Verifying => rsx! {
+            span {
+                class: "inline-flex items-center",
+                span { class: "inline-block w-3 h-3 border-2 border-current border-t-transparent rounded-full animate-spin text-muted-foreground" }
+            }
+        },
+        Nip05Status::Unknown => rsx! {},
+    }
+}
--- a/src/components/nip05_badge.rs
+++ b/src/components/nip05_badge.rs
@@ -32,7 +32,10 @@ pub fn Nip05Badge(props: Nip05BadgeProps) -> Element {
                 crate::platform::timer::sleep_ms(200).await;
                 let current = nip05::get_nip05_status(&pk, &nip);
                 if *status.read() != current {
-                    status.set(current);
+                    status.set(current.clone());
+                }
+                if matches!(&current, Nip05Status::Verified | Nip05Status::Impersonator) {
+                    break;
                 }
             }
         }
--- a/src/routes/articles/article_detail.rs
+++ b/src/routes/articles/article_detail.rs
@@ -39,12 +39,13 @@ pub fn ArticleDetail(naddr: String) -> Element {
             loading.set(true);
             error.set(None);
             match crate::utils::nip19::parse_naddr(&naddr_str) {
-                Ok((pubkey, identifier)) => {
-                    crate::stores::profiles::PROFILE_CACHE.write().pop(&pubkey);
-                    match nostr_client::fetch_event_by_coordinate(
-                        nostr_sdk::Kind::LongFormTextNote.as_u16(),
-                        pubkey.clone(),
-                        identifier,
+                Ok(parsed) => {
+                    crate::stores::profiles::PROFILE_CACHE.write().pop(&parsed.pubkey);
+                    match nostr_client::fetch_event_by_coordinate_with_relays(
+                        parsed.kind,
+                        parsed.pubkey.clone(),
+                        parsed.identifier,
+                        parsed.relay_hints,
                     )
                     .await
                     {
@@ -55,7 +56,7 @@ pub fn ArticleDetail(naddr: String) -> Element {
                             spawn(async move {
                                 profile_prefetch::prefetch_event_authors(&[event]).await;
                                 if let Some(profile) =
-                                    crate::stores::profiles::get_cached_profile(&pubkey)
+                                    crate::stores::profiles::get_cached_profile(&parsed.pubkey)
                                 {
                                     let mut metadata = nostr_sdk::Metadata::new();
                                     if let Some(name) = profile.name {
--- a/src/routes/articles/article_detail.rs
+++ b/src/routes/articles/article_detail.rs
@@ -26,20 +26,35 @@ pub fn ArticleDetail(naddr: String) -> Element {
     let mut is_liking = use_signal(|| false);
     let mut is_liked = use_signal(|| false);
     let mut like_count = use_signal(|| 0usize);
+    let mut load_generation = use_signal(|| 0u32);
     let has_signer = *nostr_client::HAS_SIGNER.read();
     let (cached_muted_posts, cached_blocked_users) = use_mute_block_cache();
-    use_effect(move || {
+
+    use_effect(use_reactive!(|naddr| {
         let naddr_str = naddr.clone();
+        let gen = load_generation.peek().wrapping_add(1);
+        load_generation.set(gen);
+
+        article.set(None);
+        loading.set(true);
+        error.set(None);
+        author_metadata.set(None);
+        comments.set(Vec::new());
+        loading_comments.set(false);
+        is_liked.set(false);
+        like_count.set(0);
+
         let client_initialized = *nostr_client::CLIENT_INITIALIZED.read();
         if !client_initialized {
             log::info!("Waiting for client initialization before loading article...");
             return;
         }
+
+        let lg = load_generation;
         spawn(async move {
-            loading.set(true);
-            error.set(None);
             match crate::utils::nip19::parse_naddr(&naddr_str) {
                 Ok(parsed) => {
+                    let author_pubkey_for_meta = parsed.pubkey.clone();
                     match nostr_client::fetch_event_by_coordinate_with_relays(
                         parsed.kind,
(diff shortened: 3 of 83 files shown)


src/stores/nostr_client/mod.rs
```diff
@@ -181,7 +181,7 @@ pub async fn initialize_client() -> std::result::Result<Arc<Client>, String> {
                 timeout: Duration::from_secs(30),
             })
             .gossip(GossipOptions::default().limits(gossip_limits))
-            .pool(RelayPoolOptions::new().max_relays(Some(50)));
+            .pool(RelayPoolOptions::new());
```
goodmorningbitcoin, line 184: Test 0

goodmorningbitcoin (commented):
Test comment 0

src/routes/articles/article_detail.rs
```diff
@@ -39,12 +39,13 @@ pub fn ArticleDetail(naddr: String) -> Element {
...
             error.set(None);
             match crate::utils::nip19::parse_naddr(&naddr_str) {
-                Ok((pubkey, identifier)) => {
-                    crate::stores::profiles::PROFILE_CACHE.write().pop(&pubkey);
-                    match nostr_client::fetch_event_by_coordinate(
-                        nostr_sdk::Kind::LongFormTextNote.as_u16(),
-                        pubkey.clone(),
-                        identifier,
+                Ok(parsed) => {
+                    crate::stores::profiles::PROFILE_CACHE.write().pop(&parsed.pubkey);
```
goodmorningbitcoin: Test 1

goodmorningbitcoin (commented):
Test comment 1

src/stores/relay/coverage.rs
```diff
@@ -381,6 +381,26 @@ pub async fn cleanup_ephemeral_relays(client: &Client, urls: &[String]) {
     }
 }
 
+/// Remove gossip-only relays from the pool.
```
goodmorningbitcoin, line 384: Test 3

goodmorningbitcoin (commented):
Test comment 3

src/utils/nip19.rs
```diff
@@ -17,23 +23,37 @@ pub fn normalize_pubkey(pubkey_str: &str) -> Result<String, String> {
...
+                .map_err(|_| format!("Invalid kind: {}", parts[0]))?;
+            Ok(ParsedNaddr {
+                kind,
+                pubkey: parts[1].to_string(),
+                identifier: parts[2].to_string(),
+                relay_hints: vec![],
+            })
         } else if parts.len() == 2 {
-            Ok((parts[0].to_string(), parts[1].to_string()))
+            Ok(ParsedNaddr {
```
goodmorningbitcoin: Test 4

goodmorningbitcoin (commented):
Test comment 4

src/stores/auth_store.rs
```diff
@@ -409,21 +409,31 @@ async fn restore_nostr_connect(
...
-    // Wait for user relay lists before starting subscriptions
-    // Critical for NIP-46 where signer restoration is slow and
-    // relay application happens concurrently in set_signer()'s spawn_forever
+    // Wait for user relay lists BEFORE any NIP-78 fetches.
+    // Without this, fetches hit bootstrap relays only and silently return
+    // empty results, which get baked in as LoadedDefaults (never retried).
     crate::stores::relay::wait_for_user_relays(
         std::time::Duration::from_secs(10),
         "run_post_login_init",
     )
```
goodmorningbitcoin, line 418: Test 5

goodmorningbitcoin (commented):
Test comment 5

src/services/aggregation/counting.rs
```diff
@@ -222,14 +222,20 @@ async fn fetch_events_for_filter(
     filter: Filter,
     timeout: Duration,
 ) -> Result<FetchEventsPage, String> {
-    let db_events: Vec<Event> = match client.database().query(filter.clone()).await {
+    let (db_events, relay_events) = tokio::join!(
```
goodmorningbitcoin: Test 6

goodmorningbitcoin (commented):
Test comment 6

goodmorningbitcoin (commented):
test

src/routes/note.rs
```diff
@@ -537,6 +597,7 @@ pub fn Note(note_id: String, from_voice: Option<String>) -> Element {
         replies.set(Vec::new());
         parent_events.set(Vec::new());
         interaction_counts.set(HashMap::new());
```
goodmorningbitcoin, line 619: test line 599

src/stores/nostr_client/mod.rs
goodmorningbitcoin, line 184: 🔴 **Critical: Relay pool cap removed without replacement guard.**

Removing `max_relays(Some(50))` from both init paths (here and line 232) prevents the `TooManyRelays` error, but doesn't address root cause — relay pool growth. Several paths add relays without guaranteed cleanup:

1. `connect_ephemeral_relays` only tracks *successfully connected* relays in `newly_added`. Failed connections stay in pool forever.
2. `fetch_event_by_coordinate_with_relays` Phase 4 adds ephemeral relays per detail page fetch.
3. Relay feeds (`home/mod.rs`) add via `add_read_relay` — cleanup may remove user-configured relays.
4. `counting.rs` uses `client.fetch_events()` (gossip mode) which can add discovered relays.

`cleanup_gossip_relays` exists in coverage.rs but is `#[allow(dead_code)]` — never called.

**Fix:** Either raise cap to `max_relays(Some(200))` as a safety net, or wire up `cleanup_gossip_relays` to run periodically (e.g., after `wait_for_user_relays`, or on a 60s timer).

src/routes/articles/article_detail.rs
goodmorningbitcoin: 🔴 **Critical: PROFILE_CACHE evicted on article navigation.**

`PROFILE_CACHE.write().pop(&parsed.pubkey)` removes the author's profile from the 5000-entry LRU cache on every visit. This cache has a 24-hour TTL and is designed to persist across navigations. Evicting entries forces relay round-trips to re-fetch the profile, causing visible flash (name/avatar disappearing briefly).

This PR fixes the same bug in `note.rs` (removes `.clear()`), but the article page still has it via `.pop()`.

**Fix:** Delete this `.pop()` call. The LRU handles eviction, the TTL handles staleness, and `fetch_profile()` already does background refresh.

src/routes/note.rs
goodmorningbitcoin, line 619: ⚠️ **Warning: Leaked interaction stream subscription on note navigation.**

When `note_id` changes (navigating Note A → Note B), this reset block clears replies, parent_events, etc. but does **not** unsubscribe the old `interaction_stream_handle`. The handle is only cleaned up in `use_drop` (line ~802), which fires on component unmount — not on prop change.

Every note navigation leaks a `stream_interaction_counts` subscription that continues receiving and processing events for the old note.

**Fix:** Add before the state resets:
```rust
if let Some(handle) = interaction_stream_handle.write().take() {
    spawn(async move { handle.unsubscribe().await; });
}
```

src/stores/relay/coverage.rs
goodmorningbitcoin, line 384: 💡 **Suggestion: `cleanup_gossip_relays` is dead code — wire it up.**

This function would reclaim pool slots from accumulated gossip relays — exactly what removing `max_relays` exacerbates. Consider calling it:
- After `wait_for_user_relays` completes in `run_post_login_init`
- On a periodic timer (e.g., every 60s)
- At the start of `fetch_event_by_coordinate_with_relays`

Or remove it entirely if the intent is to let the pool grow unbounded.

src/utils/nip19.rs
goodmorningbitcoin: ⚠️ **Warning: 2-part colon naddr sets `kind: 0` — invalid for parameterized events.**

When parsing `pubkey:identifier` (2-part format), kind is hardcoded to 0. Kind 0 is `Metadata`, not a parameterized replaceable event kind. Callers passing this to `fetch_event_by_coordinate_with_relays` will build a filter with `kind(0)` that won't match any parameterized event.

**Fix:** Return an error for 2-part format (require all 3: `kind:pubkey:identifier`), or document which callers rely on this fallback and ensure they override the kind.

src/stores/auth_store.rs
goodmorningbitcoin, line 418: 💡 **Suggestion: 10-second `wait_for_user_relays` blocks ALL NIP-78 loads.**

On slow networks (especially mobile), this gate adds up to 10 seconds of blank screen before any app data (sidebar, reactions, settings) appears. Consider:
1. Reducing the timeout (user relays typically resolve in <2s on WiFi)
2. Starting NIP-78 loads optimistically with bootstrap relays, then re-running when user relays arrive
3. Showing a loading indicator to the user during this wait

src/services/aggregation/counting.rs
goodmorningbitcoin: ⚠️ **Warning: Concurrent DB + gossip fetch per batch creates relay connection burst.**

`tokio::join!` runs DB query and `client.fetch_events()` (gossip mode) simultaneously. Combined with `join_all` for batch parallelism, this creates a thundering herd: for 500 uncached IDs with batch_size=100, that's 5 concurrent batches each doing a gossip fetch.

Each gossip fetch can discover and add relays to the pool. With `max_relays` removed, this accumulates relays rapidly.

**Consider:** Using `fetch_events_from_connected_relays` instead of `client.fetch_events()` for counting queries, since the counts don't need to be exhaustive — approximate counts from connected relays are sufficient.

goodmorningbitcoin (changes requested):
## 🤖 Hermes Agent Review

**Verdict: Changes Requested** — 2 critical, 3 warnings, 3 suggestions

### 🔴 Critical
1. **Relay pool cap removed without replacement guard** (`nostr_client/mod.rs:184,232`) — Unlimited pool growth with incomplete cleanup. `cleanup_gossip_relays` is dead code.
2. **PROFILE_CACHE evicted on article navigation** (`article_detail.rs:43`) — Same bug this PR fixes in note.rs, still present for articles

### ⚠️ Warnings
1. **Leaked interaction stream subscription** (`note.rs:599`) — Old subscription not unsubscribed on note_id change
2. **Counting gossip burst** (`counting.rs:225`) — Parallel gossip fetches accumulate relays
3. **Invalid kind=0 for 2-part naddr** (`nip19.rs:51`) — Won't match any parameterized event

### 💡 Suggestions
- `cleanup_gossip_relays` is dead code — wire it up or remove it
- 10s `wait_for_user_relays` blocks all NIP-78 loads on slow networks
- Relay feed cleanup in `home/mod.rs` may remove user-configured relays
- `INTERACTION_GLOBAL_CACHE` has no eviction — grows without bound
- Dead code: `specialty.rs::add_relays_from_strings`, `hooks/mod.rs` unused_imports

### ✅ What Looks Good
- **naddr relay hint extraction** properly wired through all detail pages — major win for #318
- **DB-first reply loading** with parallel fetch phases — significant perf improvement
- **`merge_new_replies`** with HashSet dedup — sound and avoids clone-sort cycles
- **Post-login init** properly parallelized after relay readiness gate
- **Combined filter** in `threaded_comment.rs` reduces 3 sequential fetches to 1
- **Video thumbnails** cross-platform approach with blurhash fallback
- **Stream interaction counts** for real-time updates in note detail

---
*Reviewed by Hermes Agent*

goodmorningbitcoin (commented):
## PR Review: 3 new commits (cd1cccba, 2af8ad5c, 1edc3187)

### Summary
This PR adds NIP-05 verification badges, a Zaps tab on profile pages, article nostr-URI embeds, auto-linking bare URLs, and various fixes (stream leak, cache clearing, counting path).

### Findings
- 🔴 **1 Critical** — NIP-05 polling loop never exits
- 🐛 **1 Bug** — Dead interaction_counts signal drives useless polling
- ⚠️ **3 Medium** — Zaps tab uses gossip instead of targeted fetch; zaps bypass DB cache; redundant use_effect + use_future
- 💡 **3 Low** — HTML fragment splitting edge case; missing profile fetch for zap senders; incomplete URL tag protection
- ✅ **3 Positive** — Stream unsub fix, cache clear removal, gossip avoidance in counting

Overall: Good improvements, but the infinite polling loop and dead signal need fixing before merge.

src/components/nip05_badge.rs
```diff
@@ -0,0 +1,114 @@
...
+        let pk = pubkey_for_effect.clone();
+        let nip = nip05_for_effect.clone();
+        async move {
+            loop {
+                crate::platform::timer::sleep_ms(200).await;
+                let current = nip05::get_nip05_status(&pk, &nip);
+                if *status.read() != current {
+                    status.set(current);
+                }
+            }
```
goodmorningbitcoin, line 40: 🔴 **CRITICAL: Infinite polling loop with no exit condition**

This `loop` polls every 200ms **forever** — even after reaching terminal states (`Verified`, `Impersonator`). Each component instance that mounts runs this forever.

The correct fix is to break on terminal states:
```rust
loop {
    crate::platform::timer::sleep_ms(200).await;
    let current = nip05::get_nip05_status(&pk, &nip);
    if *status.read() != current {
        status.set(current);
    }
    // Exit on terminal states
    if matches!(current, Nip05Status::Verified | Nip05Status::Impersonator) {
        break;
    }
}
```
On a profile page with many zap senders, you could have dozens of these loops running simultaneously, each doing an LRU lookup every 200ms.

src/routes/profile.rs
```diff
@@ -517,6 +544,64 @@ pub fn Profile(pubkey: String) -> Element {
         });
     };
     let sentinel_id = use_infinite_scroll(load_more, current_tab_has_more, loading_events);
+
```
goodmorningbitcoin, line 545: 🐛 **BUG: Dead signal drives useless polling loop**

The `interaction_counts` signal is declared (line 51), reset on pubkey change (line 69), and passed to this polling loop — but it is **never read** anywhere in the rendering code (rsx!). No `NoteCard` or other component reads `interaction_counts`.

This means:
- DOM queries (`document.querySelectorAll`) run every 300ms for nothing
- `fetch_counts_for_visible` fires network requests whose results are never displayed
- CPU/battery waste on mobile

Either wire up the counts to display on note cards, or remove this entire block (lines 204–242) and the `interaction_counts` signal declaration.

src/routes/profile.rs
```diff
@@ -1422,6 +1545,10 @@ fn build_tab_filter(
             .author(public_key)
             .kind(Kind::Reaction)
             .limit(limit),
```
goodmorningbitcoin, line 1505: ⚠️ **MEDIUM: Zaps tab uses gossip path instead of targeted fetch**

`fetch_events_from_relays` calls `client.fetch_events()` internally — the gossip path. This triggers relay discovery and adds discovered relays to the pool.

The profile page already resolves `user_write_relays` (lines ~102-109). The zaps tab should use `fetch_profile_events_targeted` (like other tabs) or the ephemeral relay pattern with the already-resolved write relays, not gossip.

```rust
// Should be something like:
let events = nostr_client::fetch_profile_events_targeted(
    pubkey, filter, Duration::from_secs(10)
).await
```

src/routes/profile.rs
```diff
@@ -1321,26 +1449,21 @@ fn VertsVideoCard(event: NostrEvent) -> Element {
...
-                            } else {
-                                let _ = video.pause();
-                                video.set_current_time(0.0);
-                            }
-                        }
-                    }
-                }
-            }
-            #[cfg(not(feature = "web"))]
-            let _ = (&id, hovering);
```
goodmorningbitcoin, line 1340: ⚠️ **MEDIUM: Zaps tab bypasses database cache entirely**

This always returns an empty `LoadOutcome`, completely bypassing the database cache layer. Zap receipts (kind 9735) **are** stored in the SDK database after relay fetches, so every tab load goes straight to network.

Should query the DB like other tabs:
```rust
if matches!(tab, ProfileTab::Zaps) {
    let filter = build_tab_filter(public_key, tab, until, 100);
    let events = nostr_client::fetch_profile_events_db(filter).await?;
    let processed = process_tab_events(events, tab);
    // ... compute oldest_cursor and return
}
```

src/components/nip05_badge.rs
```diff
@@ -0,0 +1,114 @@
...
+    let pubkey_for_effect = props.pubkey.clone();
+    let nip05_for_effect = props.nip05.clone();
+
+    use_effect(use_reactive(
+        (&props.pubkey, &props.nip05),
+        move |(pubkey, nip05)| {
+            let current = nip05::get_nip05_status(&pubkey, &nip05);
+            status.set(current.clone());
+            if matches!(current, Nip05Status::Unknown) {
+                nip05::verify_nip05(&pubkey, &nip05);
```
goodmorningbitcoin, line 22: ⚠️ **MEDIUM: Redundant use_effect + use_future for same data**

This component uses **both** a `use_effect` (this line, fires on prop changes) AND a `use_future` (line 33, polls every 200ms) to observe the same `NIP05_CACHE` global.

The effect fires `verify_nip05()` which spawns an async task that writes to the cache. The polling future observes the cache. This is redundant — the polling alone suffices as the observation mechanism. The effect also reads and sets `status` synchronously from cache, which the first poll iteration would also do.

Consider removing this `use_effect` and relying solely on the polling `use_future` (with the terminal-state exit fix from the other comment). The `verify_nip05` call can move into the first iteration of the polling loop when `Unknown` is detected.

src/components/article_content.rs
```diff
@@ -1,52 +1,91 @@
...
+    Html(String),
+    NostrUri(usize),
+}
+
+fn split_html_on_markers(html: &str, uri_count: usize) -> Vec<ContentSegment> {
+    let mut segments = Vec::new();
+    let mut remaining = html;
+    for idx in 0..uri_count {
+        let marker = format!("%%NOSTR_BLUE_EMBED_{}%%", idx);
+        if let Some(pos) = remaining.find(&marker) {
```
goodmorningbitcoin, line 33: 💡 **LOW: HTML fragment splitting can produce broken markup**

`split_html_on_markers` can produce broken HTML fragments. When a marker falls inside an HTML element (e.g., `<p>%%NOSTR_BLUE_EMBED_0%%</p>`), the split produces an unclosed `<p>` before the marker and a dangling `</p>` after. While browsers auto-fix this via innerHTML, it can cause unexpected spacing/rendering artifacts.

Consider stripping empty HTML wrappers around markers, or splitting at block boundaries instead of at arbitrary positions.

src/routes/profile.rs
```diff
@@ -1968,6 +2106,28 @@ async fn load_tab_events(
...
+            let oldest_cursor = event_vec.last().map(|e| e.created_at.as_secs());
+            Ok(LoadOutcome {
+                events: event_vec,
+                oldest_cursor,
+                relay_count,
+            })
+        }
     }
 }
 fn get_display_name(metadata: &nostr_sdk::Metadata, pubkey: &str) -> String {
```
goodmorningbitcoin, line 2094: 💡 **LOW: ZapEntryCard doesn't trigger profile fetch for unknown senders**

`profiles::get_profile(&hex)` is a synchronous LRU cache read. If the sender's profile isn't cached (common for zap senders who aren't in your social graph), the card shows truncated hex instead of a name, and **no background fetch is triggered** to populate the cache.

Consider calling `profiles::fetch_profile(&hex)` instead (or in addition) to trigger a relay fetch, so the card eventually populates with the sender's name/avatar:
```rust
spawn(async move {
    let metadata = profiles::get_profile(&hex)
        .or_else(|| profiles::fetch_profile(&hex).await);
    sp.set(metadata);
});
```

src/utils/parsing/markdown.rs
```diff
@@ -14,7 +21,60 @@ pub fn render_markdown(markdown: &str) -> String {
...
+    let linked = auto_link_bare_urls(&sanitized);
+    convert_mermaid_blocks(&linked)
+}
+
+fn auto_link_bare_urls(html: &str) -> String {
+    static A_TAG: Lazy<Regex> =
+        Lazy::new(|| Regex::new(r"<a\s[^>]*>.*?</a>").expect("a tag regex"));
+    static IMG_TAG: Lazy<Regex> =
+        Lazy::new(|| Regex::new(r"<img\s[^>]*/?>").expect("img tag regex"));
+    static BARE_URL: Lazy<Regex> =
```
goodmorningbitcoin, line 33: 💡 **LOW: auto_link_bare_urls only protects `<a>` and `` tags**

Only `<a>` and `` tags are protected from double-linking. Other tags with URL attributes (`<source src>`, `<video src>`, `<iframe src>`) are not protected. Bare URLs inside these attributes could get double-wrapped in `<a>` tags.

Low risk for user-generated markdown content since these tags are unlikely (and ammonia sanitization probably strips them anyway), but worth noting for robustness. If ammonia strips them pre-linking, this is a non-issue.

src/routes/note.rs
```diff
@@ -8,7 +8,8 @@ use crate::components::{ClientInitializing, EditProposalCard, NoteCard, Threaded
 use crate::hooks::{use_mute_block_cache, use_relay_subscription};
 use crate::routes::Route;
 use crate::services::aggregation::{
-    fetch_interaction_counts_batch, InteractionCounts,
+    fetch_interaction_counts_batch, fetch_local_db_counts, stream_interaction_counts,
+    InteractionCounts, InteractionStreamHandle,
 };
 use crate::stores::back_navigation;
 use crate::stores::nostr_client;
```
goodmorningbitcoin, line 15: ✅ **Good fix** — The interaction stream handle is now properly unsubscribed on re-navigation via `.take()` + `spawn(async { handle.unsubscribe() })`. This addresses the stream subscription leak flagged in the previous review. Nice.

src/routes/video/video_detail.rs
```diff
@@ -13,10 +13,6 @@ use dioxus::prelude::*;
 use nostr_sdk::prelude::*;
 use nostr_sdk::{Event, EventId, PublicKey};
 use std::time::Duration;
-#[cfg(feature = "web")]
-use wasm_bindgen::JsCast;
-#[cfg(feature = "web")]
-use web_sys::HtmlVideoElement;
 #[derive(Clone, Copy, PartialEq, Debug)]
 pub(crate) enum FeedType {
```
goodmorningbitcoin, line 17: ✅ **Good fix** — Removed `PROFILE_CACHE.write().clear()` from video detail page navigation. The LRU + TTL already handles eviction and staleness, and the `clear()` was a major contributor to slow comment rendering (every video navigation evicted all cached profiles).

src/services/aggregation/counting.rs
```diff
@@ -218,19 +218,25 @@ struct FetchEventsPage {
 }
 
 async fn fetch_events_for_filter(
-    client: &Client,
+    client: &std::sync::Arc<Client>,
     filter: Filter,
     timeout: Duration,
 ) -> Result<FetchEventsPage, String> {
-    let db_events: Vec<Event> = match client.database().query(filter.clone()).await {
```
goodmorningbitcoin, line 225: ✅ **Good change** — Switched from `client.fetch_events()` (gossip path) to `fetch_events_from_connected_relays_with_client()` for interaction counting. This avoids gossip relay discovery and pool bloat in the counting hot path. This is the right pattern for high-frequency internal queries.

goodmorningbitcoin:
## Review: Round 3 (commits ad92805..8488af5)

### ✅ Previous Review Fixes — All Confirmed

All 4 critical/bug issues from Round 2 were fixed correctly:

1. **NIP-05 infinite polling** → Now breaks on `Verified | Impersonator` terminal states ✅
2. **Dead `interaction_counts` loop** → Completely removed from profile.rs ✅
3. **Zaps tab gossip path** → Switched to `fetch_profile_events_targeted` ✅
4. **ZapEntryCard LRU-only read** → Now calls `fetch_profile()` as fallback ✅

---

### 🔴 CRITICAL

**1. `get_cached_replies` does full linear scan of 10K LRU cache**
`src/stores/ndb/mod.rs`

```rust
pub fn get_cached_replies(event_id: &EventId, kinds: &[Kind]) -> Vec<nostr::Event> {
    cache.iter()
        .filter(|(_, event)| {
            kinds.contains(&event.kind) &&
            event.tags.iter().any(|tag| tag.content() == Some(event_id_hex))
        })
        .map(|(_, event)| event.clone())
        .collect()
}
```

Called from `fetch_replies_db()` on every note page load. 10K entries × ~5 tags each = ~50K string comparisons per page visit. On desktop this is fast enough, but architecturally wrong for a hot path. Consider:
- A secondary index: `HashMap<EventId, Vec<EventId>>` for parent→replies
- Or cap the scan and bail after N iterations

---

### ⚠️ MEDIUM

**2. `fetch_event_targeted` logs every phase at `info!` level**
`src/stores/nostr_client/fetching.rs`

Every phase (0–5) now logs at `info!` with the full event ID hex. This is the hot path — called for every note click, every embedded card render. Will produce massive log output drowning useful signal. Should be `debug!` or `trace!`.

**3. `EVENT_BRIDGE_CACHE` Mutex contention on native**
`src/stores/ndb/mod.rs`

`cache_event()` is called from notification_dispatcher, counting.rs, fetching.rs, and note.rs. All lock a `std::sync::Mutex` on the same 10K LRU. If nostrdb ingestion is concurrent, this creates contention on the event processing pipeline. Consider `parking_lot::Mutex` or a lock-free approach.

**4. `set_read_only` resets `USER_RELAYS_APPLIED` but old relays stay in pool**
`src/stores/nostr_client/mod.rs`

```rust
*relay::USER_RELAYS_APPLIED.write() = false;
```

Resets the flag on logout, good. But the user's NIP-65 relays remain in the pool. On login as a different user, `init_user_relay_lists` adds NEW relays on top of the old ones. Combined with no `max_relays` cap, this causes unbounded pool growth across account switches.

---

### 💡 LOW

**5. `referrerpolicy: "no-referrer"` blanket addition across ~15 image tags**

Good for privacy. But some CDN hosts use Referer for hotlink protection and return 403. Consider `"no-referrer-when-downgrade"` (only strips on HTTPS→HTTP).

**6. `FALLBACK_ART_URL` duplicated 3 times**

Same constant in `player_bar.rs`, `player_expanded.rs`, and `unified_track_card.rs`. Should be a shared constant.

---

### ✅ Positive Changes

**7. Feed merge with `merge_feed_items` instead of replace** — Pagination now merges new items into existing ones instead of replacing the entire feed. Fixes feed jumps/flickers on scroll.

**8. Cleanup on unmount for Home component** — Unsubscribes real-time subscriptions, ndb subscriptions, interaction streams, and cleans up ephemeral relays. Was a resource leak.

**9. `upgrade_to_https` for podcast images** — Podcast RSS feeds commonly use `http://` image URLs which fail on HTTPS pages. Tests included.

**10. Cross-platform scroll-to-top** — Replaced 6 `web_sys::window().scroll_to()` instances with `set_scroll_y(0.0)` working on both web and native. Clean `#[cfg]` removal.

**11. NDB bridge cache + `event_by_id` optimization** — `fetch_event_targeted` now uses `event_by_id` (direct PK lookup) instead of `database().query(filter)` (filter pipeline). Bridge cache sits above DB check for the nostrdb async ingestion gap. Both solid improvements.

**12. `AlreadyBorrowed` crash fix** — `show_zap_dialog_for_track` and `sync_native_playback_snapshot` correctly separate `.peek().clone()` from `.write()` to avoid double-borrowing the same signal.

PR #328, opened 2026-05-25
Release 0.8.15: NIP-64 Chess, NIP-29 Groups, Address Viewer, Weather Pages

patrickulrich:
New Features
- NIP-64 Chess — Full chess integration with live play via kind 30 events (Jester protocol) and PGN records via kind 64. Includes real-time move subscriptions, resign/draw, SAN replay, FEN desync recovery, auto game-end + PGN publishing, and a dedicated chess lobby with open challenges, active games, and completed games list.
- NIP-29 Relay Groups — Community/group support with routes, components, store, and sidebar integration.
- NIP-50 Search Overhaul — Query parser with engagement scoring and search history.
- Unified Address Viewer — Single /:naddr route that dispatches to 37 viewer components (articles, polls, calendars, livestreams, podcasts, chess, communities, code repos, and more).
- Weather Dashboard — New weather page with open-source forecast data.
- NIP-66 Relay Explorer — Discovery and browsing of Nostr relays.
- NIP-70 Protected Events — Support for protected event handling.

Media & Performance
- MPRIS / Media Session — Desktop media controls, queue persistence, media player bar visibility across sessions
- Feed Virtualization — Scroll restore, optimistic publishing, draft debounce, stale guard hook
- Offline Banner — Connection status indicator


src/components/chess/board.rs
```diff
@@ -0,0 +1,325 @@
...
+            }
+        }
+    }
+}
+
+#[allow(clippy::too_many_arguments)]
+fn apply_move(
+    src: SquareCoord,
+    dst: SquareCoord,
+    _promo: Option<PromotionPiece>,
```
goodmorningbitcoin: 🔴 **Critical: Promotion piece is completely ignored.**

The `_promo` parameter is unused in `apply_move`. On line 303, `get_legal_move(src.file, src.rank, dst.file, dst.rank)` is called without the promotion piece. When a pawn reaches the back rank, there are up to 4 legal moves (Queen/Rook/Bishop/Knight) with identical from/to squares. `.find()` returns the first match — likely always Queen — regardless of the user's selection in the promotion dialog.

**Fix:** Pass the promotion piece to `get_legal_move` or use a different API that disambiguates promotion moves by piece type.

src/routes/games/chess/game.rs
```diff
@@ -0,0 +1,795 @@
...
+                }
+            });
+        });
+
+    let perspective = match *viewer_role.read() {
+        ViewerRole::BlackPlayer => rschess::Color::Black,
+        _ => rschess::Color::White,
+    };
+
+    let white_pubkey = {
```
goodmorningbitcoin, line 323: 🔴 **Critical: `white_pubkey` and `black_pubkey` have identical match arms — both resolve to the same player.**

Both blocks match on `ViewerRole::BlackPlayer => my_pubkey` and `ViewerRole::WhitePlayer => opp`. This means `white_pubkey` and `black_pubkey` are always equal, so White and Black names are always the same.

For `white_pubkey`, the match should be:
- `WhitePlayer => my_pubkey`
- `BlackPlayer => opp`

And `black_pubkey` should be the inverse. Currently the display shows the wrong player on each side of the board.

src/stores/chess/lobby_state.rs
```diff
@@ -0,0 +1,105 @@
...
+            self.active_games.push(ActiveGame {
+                game_id: *game_id,
+                start_event_id,
+                white_pubkey: white_pk,
+                black_pubkey: black_pk,
+                viewer_role,
+                move_count: 0,
+                last_move_san: None,
+                is_my_turn: challenge.challenger_color == rschess::Color::White,
+                last_move_at: std::time::SystemTime::now()
```
goodmorningbitcoin: 🔴 **Critical: `std::time::SystemTime::now()` panics on WASM.**

This code path runs when a user accepts a chess challenge from the lobby UI. On WASM targets, `SystemTime::now()` is not supported and will panic, crashing the app.

**Fix:** Use `crate::platform::timestamp::now_secs()` or `web_time::SystemTime::now()` with a cfg gate, consistent with the project's WASM compatibility patterns.

src/stores/social/group_store/fetch.rs
```diff
@@ -0,0 +1,346 @@
...
+    if let Some(event) = events.into_iter().next() {
+        let groups = parse_groups_list_from_event(&event);
+        return Ok(groups);
+    }
+    Ok(Vec::new())
+}
+
+pub async fn fetch_user_groups() -> std::result::Result<Vec<Group>, String> {
+    *GROUPS_LOADING.write() = true;
+    let group_list = fetch_user_groups_list().await?;
```
goodmorningbitcoin: 🔴 **Critical: `GROUPS_LOADING` stuck as `true` on fetch failure.**

If `fetch_user_groups_list()` on line 226 returns `Err`, the `?` operator causes an early return. Lines 235-236 (`GROUPS_LOADING = false`, `GROUP_INITIALIZED = true`) are never reached. The UI will permanently show a loading spinner and the groups list becomes unusable until the app restarts.

**Fix:** Wrap in a match or use a `finally`-style block to ensure loading state is always reset:
```rust
let result = fetch_user_groups_list().await;
*GROUPS_LOADING.write() = false;
let group_list = result?;
```

src/components/groups/settings.rs
```diff
@@ -0,0 +1,375 @@
...
+            },
+            onclick: {
+                let relay = relay_url.clone();
+                let gid = group_id.clone();
+                let f = field.clone();
+                move |_| {
+                    let relay = relay.clone();
+                    let gid = gid.clone();
+                    let f = f.clone();
+                    spawn(async move {
```
goodmorningbitcoin, line 353: 🔴 **Critical: `StatusToggle` can only ENABLE status flags, never disable them.**

The `onclick` handler always passes `Some(true)` for the selected field and `None` for others. There's no toggle logic — clicking an already-active button sends the same `Some(true)` value again (a no-op). The `edit_group_status` API only adds tags for `Some(true)` values, so there's no mechanism to remove a status.

**Fix:** Toggle the `active` state:
```rust
let new_val = if active { None } else { Some(true) };
// Then use new_val for the matching field
```

goodmorningbitcoin (commented):
Batch 0 test

src/routes/games/chess/game.rs
```diff
@@ -0,0 +1,795 @@
...
+                    game_title.set("Spectating".to_string());
+                }
+                is_loading.set(false);
+
+                if my_pubkey.is_none() || event_id.is_none() {
+                    return;
+                }
+                let eid = event_id.unwrap();
+
+                loop {
```
goodmorningbitcoin, line 295: ⚠️ **Warning: Infinite polling loop with no cancellation.**

The `loop { sleep(POLL_INTERVAL_SECS); fetch... }` runs forever with no break condition or cancellation mechanism. If the component unmounts or the `use_effect` re-runs (e.g., `CLIENT_INITIALIZED` signal changes), a new loop starts while the old one continues — causing duplicate concurrent polling and resource leaks.

**Fix:** Use a cancellation token (e.g., a `Signal<bool>` that the loop checks each iteration) or store the `JoinHandle` and abort it in the effect's cleanup.

src/stores/chess/game_state.rs
```diff
@@ -0,0 +1,349 @@
...
+            .map_err(|e| format!("Invalid PGN: {:?}", e))?;
+        let tags: Vec<(String, String)> = pgn.tag_pairs().iter()
+            .map(|(k, v)| (k.clone(), v.clone()))
+            .collect();
+        let board = pgn.board().clone();
+        let movetext = board.gen_movetext();
+        let san_list: Vec<String> = if movetext.is_empty() {
+            vec![]
+        } else {
+            let re = regex::Regex::new(r"\d+\.\s*|\d+\.\.\.\s*").unwrap();
```
goodmorningbitcoin: ⚠️ **Warning: Regex compiled on every `from_pgn()` call — expensive on WASM.**

`regex::Regex::new()` compiles the pattern each invocation. This is CPU-expensive and the `regex` crate also adds significant WASM binary size.

**Fix:** Compile once with `std::sync::OnceLock`:
```rust
static RE: OnceLock<Regex> = OnceLock::new();
let re = RE.get_or_init(|| Regex::new(r"\d+\.\s*|\d+\.\.\.\s*").unwrap());
```
Or consider a simpler string split if the pattern is straightforward.

src/routes/games/chess/home.rs
```diff
@@ -0,0 +1,554 @@
...
+                                    ViewerRole::BlackPlayer
+                                },
+                                move_count,
+                                last_move_san: best_move
+                                    .and_then(|e| JesterContent::parse(&e.content))
+                                    .and_then(|c| c.mv),
+                                is_my_turn,
+                                last_move_at: best_move.map(|e| e.created_at.as_secs()).unwrap_or(start_ev.created_at.as_secs()),
+                            });
+                        }
```
goodmorningbitcoin, line 285: ⚠️ **Warning: `CHESS_LOBBY.write()` called 5 times in sequence.**

Each `.write()` acquires/releases the write lock independently. Between writes, a reader could observe partially updated state (e.g., new challenges but stale `active_games`).

**Fix:** Acquire a single write guard and set all fields:
```rust
let mut lobby = CHESS_LOBBY.write();
lobby.challenges = challenges;
lobby.active_games = active;
lobby.completed_games = completed;
lobby.is_loading = false;
```

src/routes/weather/weather_home.rs
```diff
@@ -0,0 +1,327 @@
...
+    _refetch_trigger: Signal<u32>,
+) -> Element {
+    let today = data.daily.first().cloned();
+    let current_hourly: Vec<crate::services::weather::HourlyForecast> = {
+        let now_secs = crate::platform::timestamp::now_secs();
+        let cutoff = now_secs.saturating_sub(3600) as i64;
+        data.hourly
+            .iter()
+            .filter(|h| {
+                h.time.parse::<i64>().ok().or_else(|| {
```
goodmorningbitcoin: ⚠️ **Warning: Current-hour filtering is effectively broken.**

The logic tries `h.time.parse::<i64>()` first (always fails — Open-Meteo returns ISO strings like `"2025-01-15T06:00"`, not unix timestamps), then falls through to `chrono::DateTime::parse_from_rfc3339(&format!("{}:00", h.time))`. But RFC 3339 requires a timezone offset, and `"2025-01-15T06:00:00"` has none. The parse fails silently, `or_else` returns `None`, and `unwrap_or(0)` produces epoch 1970 — so ALL hourly data passes the `> cutoff` filter.

**Fix:** Use `chrono::NaiveDateTime::parse_from_str` with format `"%Y-%m-%dT%H:%M"`, or use Open-Meteo's `utc_offset_seconds` to construct proper timestamps.

src/platform/mpris.rs
```diff
@@ -0,0 +1,285 @@
...
+        artwork_url: Option<&str>,
+    ) {
+        setup();
+        if let Ok(mut s) = state().lock() {
+            let mut b = Metadata::builder()
+                .title(title)
+                .artist([artist])
+                .album(album)
+                .length(Time::from_micros((duration * 1e6) as i64));
+            if let Some(art) = artwork_url {
```
goodmorningbitcoin, line 234: ⚠️ **Warning: Artwork URL construction is broken for HTTP URLs.**

When `art` is not an absolute path (i.e., it's an HTTP URL from a blossom server), the code constructs `file://{current_dir}/{art}`, producing nonsense like `file:///home/user/https://blossom.example.com/image.jpg`.

**Fix:** Check for HTTP(S) URLs and use them directly:
```rust
if art.starts_with('/') {
    format!("file://{art}")
} else if art.starts_with("http") {
    art.to_string()
} else {
    format!("file://{}/{}", std::env::current_dir()...)
}
```

goodmorningbitcoin (commented):
Batch 1 test

src/utils/media_session.rs
```diff
@@ -0,0 +1,86 @@
+#[cfg(feature = "web")]
+pub async fn update_metadata(
+    title: &str,
+    artist: &str,
+    album: &str,
+    artwork_url: Option<&str>,
+) {
+    let artwork_js = match artwork_url {
+        Some(url) => format!(
```
goodmorningbitcoin: ⚠️ **Warning: Artwork URL is hand-escaped into a JS string literal.**

The `format!` only escapes single quotes, but URLs containing backticks, backslashes, or newlines would break the JS string. While `title`/`artist`/`album` are properly serialized via `serde_json::to_string`, the artwork array is constructed manually.

**Fix:** Use `serde_json::to_string` for the artwork URL as well, or at minimum also escape backslashes and backticks.

src/utils/nip66.rs
```diff
@@ -0,0 +1,201 @@
...
+        .map(|s| s.to_string())
+        .collect();
+
+    let accepted_kinds: Vec<(u64, bool)> = event
+        .tags
+        .filter(TagKind::single_letter(Alphabet::K, true))
+        .filter_map(|t| {
+            t.content().map(|v| {
+                let (val, neg) = parse_negated(v);
+                (val.parse::<u64>().unwrap_or(0), neg)
```
goodmorningbitcoin: ⚠️ **Warning: Unparseable NIP-66 kind defaults to 0 (metadata).**

`val.parse::<u64>().unwrap_or(0)` means an unparseable kind like `!abc` maps to kind 0 (set-metadata). This could cause the relay explorer to incorrectly display relay requirements. Kind 0 is a real, commonly-used event type.

**Fix:** Skip invalid entries:
```rust
if let Ok(kind) = val.parse::<u64>() {
    // process
}
```

src/routes/relay_explorer.rs
```diff
@@ -0,0 +1,239 @@
+use crate::components::RelayDiscoveryCard;
+use crate::routes::Route;
+use crate::stores::nostr_client;
+use crate::utils::nip66;
+use dioxus::prelude::*;
+use std::time::Duration;
```
goodmorningbitcoin, line 6: ⚠️ **Warning: `std::time::Duration` may panic on WASM.**

While `Duration::from_secs()` construction itself is safe, if this route is reachable on WASM targets, any downstream code that uses `Instant` internally could panic. Verify this route is properly gated behind a native-only feature or that the Duration values are only passed to WASM-safe APIs.

src/services/search/engagement_fetch.rs
```diff
@@ -0,0 +1,149 @@
...
+        let filter = Filter::new()
+            .kinds([Kind::Reaction, Kind::Repost, Kind::from(9735)])
+            .custom_tags(
+                SingleLetterTag::lowercase(Alphabet::E),
+                id_hexes.iter().map(|s| s.as_str()),
+            )
+            .limit(500);
+
+        match client
+            .fetch_events(filter, Duration::from_secs(ENGAGEMENT_TIMEOUT_SECS))
```
goodmorningbitcoin, line 52: ⚠️ **Warning: `Duration::from_secs` passed to nostr-sdk `fetch_events` on potentially WASM targets.**

Same concern as `relay_explorer.rs` — verify the timeout is handled WASM-safely by nostr-sdk internally, or ensure this module is only compiled for native targets.

src/services/search/engagement_fetch.rs
```diff
@@ -0,0 +1,149 @@
...
+                if let Ok(msat) = amount_str.parse::<u64>() {
+                    return msat;
+                }
+            }
+        }
+    }
+    0
+}
+
+fn decode_bolt11_amount(invoice: &str) -> u64 {
```
goodmorningbitcoin: ⚠️ **Warning: Fragile BOLT11 invoice parsing.**

`decode_bolt11_amount` finds the first `'1'` character to locate the amount field. A malformed invoice with `'1'` early in the string could trick the parser. The implementation also doesn't handle multiple multiplier characters. Consider using a proper BOLT11 decoder library (e.g., `lightning-invoice`) for robustness.

goodmorningbitcoin (commented):
Batch 2 test

goodmorningbitcoin (commented):
Batch 3 test

src/services/search/query_parser.rs
```diff
@@ -0,0 +1,529 @@
...
+fn parse_time_value(input: &str) -> Option<Timestamp> {
+    if let Some(ts) = parse_relative_time(input) {
+        return Some(ts);
+    }
+    parse_date(input)
+}
+
+fn parse_relative_time(input: &str) -> Option<Timestamp> {
+    let input_lower = input.to_lowercase();
+    let now_secs = Timestamp::now().as_secs();
```
goodmorningbitcoin, line 250: ⚠️ **Warning: `Timestamp::now()` may panic on WASM.**

`Timestamp::now()` internally calls `SystemTime::now()` which panics on WASM. If this query parser is used on the web target for `since:7d` style search queries, it will crash.

**Fix:** Use `crate::platform::timestamp::now_secs()` and construct the `Timestamp` from the resulting u64.

src/stores/social/group_store/fetch.rs
```diff
@@ -0,0 +1,346 @@
...
+    let client = ensure_group_relay(relay_url).await?;
+    let url = RelayUrl::parse(relay_url).map_err(|e| format!("Invalid relay URL: {}", e))?;
+    let pubkey =
+        PublicKey::from_hex(user_pubkey).map_err(|e| format!("Invalid pubkey: {}", e))?;
+    let filter = Filter::new()
+        .kinds(vec![
+            Kind::Custom(KIND_PUT_USER),
+            Kind::Custom(KIND_REMOVE_USER),
+        ])
+        .author(pubkey)
```
goodmorningbitcoin: ⚠️ **Warning: Membership filter uses `.author(pubkey)` but NIP-28 management events are authored by admins.**

The filter for checking if a user is in a group filters for events authored BY the user. But NIP-28 `kind 9000` (PUT_USER) and `kind 9001` (REMOVE_USER) events are authored by group admins, not the user being added. This means the filter rarely returns results for regular members who were added by an admin, causing incorrect `NotInGroup` status for valid members.

**Fix:** Filter by `p-tag` containing the user's pubkey instead of by `author`.

src/components/groups/chat.rs
```diff
@@ -0,0 +1,384 @@
...
+}
+
+#[component]
+pub fn GroupChatComposer(
+    relay_url: String,
+    group_id: String,
+    reply_to: Signal<Option<GroupMessage>>,
+    editing: Signal<Option<GroupMessage>>,
+) -> Element {
+    let mut content = use_signal(String::new);
```
goodmorningbitcoin, line 273: ⚠️ **Warning: Editing content may overwrite user input.**

The `use_effect` sets `set_content(editing)` when the `editing` prop changes. If the user has already started typing before the effect runs, their input gets replaced. Check if content is already non-empty before overwriting.

src/components/groups/chat.rs
```diff
@@ -0,0 +1,384 @@
...
+            }
+            div { class: "flex gap-2",
+                input {
+                    class: "flex-1 px-3 py-2 bg-background border border-border rounded-lg text-foreground placeholder:text-muted-foreground focus:outline-none focus:ring-2 focus:ring-primary text-sm",
+                    r#type: "text",
+                    placeholder: if is_editing { "Edit message..." } else { "Message..." },
+                    value: "{content}",
+                    oninput: move |e| content.set(e.value()),
+                    onkeypress: move |e: KeyboardEvent| {
+                        if e.key() == Key::Enter && !content().is_empty() && !*sending.read() {
```
goodmorningbitcoin: ⚠️ **Warning: Duplicate send logic between `onkeypress` (Enter) and `onclick` (Send button).**

Both handlers contain an identical ~25-line block for editing vs. new message sending. This duplication risks divergence if one path is updated without the other. Extract into a shared `send_message()` function.

src/hooks/use_group_subscription.rs
```diff
@@ -0,0 +1,195 @@
...
+}
+
+#[allow(clippy::arc_with_non_send_sync)]
+fn spawn_group_fallback_listener(
+    client: Arc<nostr_sdk::Client>,
+    sub_id: SubscriptionId,
+    on_event: OnGroupEvent,
+) {
+    spawn(async move {
+        let mut notifications = client.notifications();
```
goodmorningbitcoin, line 44: ⚠️ **Warning: Fallback listener iterates ALL relay pool notifications.**

`spawn_group_fallback_listener` subscribes to `notifications()` which receives events from ALL relays and subscriptions, filtering only by `subscription_id` match in the hot loop. On a busy client with many subscriptions, this is wasteful. Consider a more targeted subscription mechanism.

goodmorningbitcoin (commented):
Batch 4 test

src/stores/chess/publish.rs
```diff
@@ -0,0 +1,133 @@
...
+        format!("Chess challenge (plays {})", color_label)
+    } else {
+        format!("Open chess challenge (plays {})", color_label)
+    };
+    tags.push(Tag::alt(alt));
+
+    let builder = EventBuilder::new(Kind::Custom(KIND_JESTER), content.to_json()).tags(tags);
+    let event = crate::stores::publish_queue::signing::sign_event_builder(builder).await?;
+    let event_id = event.id;
+    let _ = publish_queue::enqueue(
```
goodmorningbitcoin, line 49: 💡 **Suggestion: Publish errors silently discarded.**

`let _ = publish_queue::enqueue(...).await;` discards the result on all 4 publish functions (challenge, move, game_end, pgn). If publishing fails, the function still returns `Ok(event_id)`, leading the user to a game page that references an unpublished event.

Consider propagating the error or at least logging the failure.

src/components/groups/settings.rs
```diff
@@ -0,0 +1,375 @@
...
+                    }
+                    if let Some(code) = generated_code.read().as_ref() {
+                        div { class: "px-3 py-2 bg-accent rounded-lg text-sm text-foreground font-mono select-all",
+                            "{code}"
+                        }
+                    }
+                }
+            }
+
+            if can_delete {
```
goodmorningbitcoin, line 303: 💡 **Suggestion: Delete Group button has no confirmation dialog.**

A single click triggers the delete request. Accidental deletions are irreversible in Nostr (kind 9005 events can't be undone if the relay has already propagated). Should add a confirmation step similar to the "leave group" pattern used elsewhere.

src/components/groups/share.rs
```diff
@@ -0,0 +1,94 @@
...
+                            value: "{share_uri}",
+                        }
+                        button {
+                            class: "px-3 py-2 bg-primary text-primary-foreground rounded-lg text-sm shrink-0",
+                            onclick: {
+                                let uri = share_uri.clone();
+                                move |_| {
+                                    let uri = uri.clone();
+                                    spawn(async move {
+                                        #[cfg(all(feature = "web", target_arch = "wasm32"))]
```
goodmorningbitcoin, line 66: 💡 **Suggestion: Web Share API fallback is a no-op.**

The `#[cfg(all(feature = "web", target_arch = "wasm32"))]` branch uses the Web Share API, but the fallback (line 73+) only logs — no copy-to-clipboard alternative. Users on WASM browsers without Web Share API support will click the button and nothing happens.

src/stores/weather/weather_store.rs
```diff
@@ -0,0 +1,37 @@
...
+use super::location_store;
+use crate::services::weather::{fetch_all_weather, WeatherData};
+
+pub static WEATHER_DATA: GlobalSignal<std::collections::HashMap<String, WeatherData>> =
+    Signal::global(std::collections::HashMap::new);
+pub static WEATHER_LOADING: GlobalSignal<bool> = Signal::global(|| false);
+pub static WEATHER_ERROR: GlobalSignal<Option<String>> = Signal::global(|| None);
+
+pub async fn fetch_weather_for_current_location() -> Result<(), String> {
+    let loc = location_store::get_selected().ok_or("No location selected")?;
```
goodmorningbitcoin, line 12: 💡 **Suggestion: Unbounded `WEATHER_DATA` HashMap.**

Each `WeatherData` contains ~384 hourly entries, AQ data, and minutely data. If a user adds many locations, memory grows without limit and is never evicted. Consider adding a max location count or LRU eviction.

src/stores/ui/online_status.rs
```diff
@@ -0,0 +1,46 @@
...
+    #[cfg(feature = "web")]
+    {
+        let _ = document::eval(r#"
+            if (!window.__nostrBlueOnlineReady) {
+                window.addEventListener("online", () => { window.__nostrBlueOnline = true; });
+                window.addEventListener("offline", () => { window.__nostrBlueOnline = false; });
+                window.__nostrBlueOnlineReady = true;
+                window.__nostrBlueOnline = navigator.onLine;
+            }
+        "#).await;
```
goodmorningbitcoin, line 15: 💡 **Suggestion: Online status polls via `document::eval` every 5 seconds.**

This creates a JS→WASM bridge round-trip every 5 seconds even when status hasn't changed. Consider using the browser's native `online`/`offline` events via `wasm_bindgen` for event-driven updates instead of polling.

goodmorningbitcoin (commented):
Isolating batch 5 item 1

src/stores/social/group_store/fetch.rs
```diff
@@ -0,0 +1,346 @@
...
+        return Ok(groups);
+    }
+    Ok(Vec::new())
+}
+
+pub async fn fetch_user_groups() -> std::result::Result<Vec<Group>, String> {
+    *GROUPS_LOADING.write() = true;
+    let group_list = fetch_user_groups_list().await?;
+    let mut groups = Vec::new();
+    for (relay_url, group_id) in &group_list {
```
goodmorningbitcoin: 💡 **Suggestion: Sequential group fetching is slow.**

Each group is fetched one at a time in a for-loop. For users in many groups across different relays, this compounds latency. Use `futures::future::join_all` to fetch groups concurrently.

src/components/music/player/mod.rs
```diff
@@ -597,6 +679,10 @@ pub fn PersistentMusicPlayer() -> Element {
                             let duration = audio.duration();
                             if !duration.is_nan() {
```
goodmorningbitcoin, line 680: 💡 **Suggestion: MPRIS polling loop runs every 100ms even when idle.**

The `sleep_ms(100)` + `poll_event()` loop generates continuous CPU wakeups and position updates even when no media is playing. Consider increasing the poll interval when idle or using an event-driven approach.

goodmorningbitcoin (commented):
Isolating batch 5 item 2

goodmorningbitcoin:
💡 **Suggestion: `Shorts` → `Verts` rename invalidates existing caches** (`src/stores/feed_cache.rs:82`)

The cache key format changes from `shorts:` / `shorts_global` to `verts:` / `verts_global`. Users upgrading will have orphaned entries in IndexedDB/storage that are never cleaned up. Consider adding a migration or cleanup step for old keys.

---
*Reviewed by [Hermes Agent](https://github.com/apps/hermes-agent)* 🤖

goodmorningbitcoin:
## 🤖 Hermes Agent Code Review — PR #328

### Summary
Thorough review of **351 files** (+40,469 / -21,674 lines) across Chess (NIP-64), Groups (NIP-28), Address Viewer, Weather, Search, MPRIS, and more.

### Findings: 5 Critical · 15 Warnings · 10 Suggestions

---

**🔴 Critical Issues:**

1. **Chess promotion piece ignored** (`src/components/chess/board.rs:295`) — `_promo` parameter is unused in `apply_move`. `get_legal_move` uses `.find()` which returns the first of up to 4 promotion moves (Queen/Rook/Bishop/Knight) with identical from/to squares. The user's promotion dialog selection is silently discarded.

2. **White/black pubkeys swapped** (`src/routes/games/chess/game.rs:315`) — Both `white_pubkey` and `black_pubkey` blocks have identical match arms (`BlackPlayer => my_pubkey, WhitePlayer => opp`), causing both to always resolve to the same player. White and Black labels always show the same name.

3. **WASM panic in chess lobby** (`src/stores/chess/lobby_state.rs:98`) — `std::time::SystemTime::now()` panics on WASM targets. This runs when a user accepts a challenge from the lobby UI, crashing the app on web.

4. **Groups loading state stuck on failure** (`src/stores/social/group_store/fetch.rs:226`) — If `fetch_user_groups_list()` fails, the `?` operator returns early before resetting `GROUPS_LOADING = false`. The UI permanently shows a loading spinner until app restart.

5. **Group status can only enable, never disable** (`src/components/groups/settings.rs:359`) — `StatusToggle` always passes `Some(true)` for the selected field. Clicking an already-active button is a no-op. There is no mechanism to remove a status flag.

---

**⚠️ Warnings:**

| # | File | Issue |
|---|------|-------|
| 1 | `game.rs:288` | Infinite polling loop with no cancellation — resource leak on component unmount |
| 2 | `game_state.rs:61` | Regex compiled on every `from_pgn()` call — expensive on WASM, use `OnceLock` |
| 3 | `home.rs:285` | `CHESS_LOBBY.write()` called 5× sequentially — readers see partial state |
| 4 | `weather_home.rs:139` | Hourly time filter broken — RFC 3339 parse fails for Open-Meteo timestamps, all data passes |
| 5 | `mpris.rs:234` | Artwork URL broken for HTTP URLs — prepends `file://` + cwd to blossom URLs |
| 6 | `media_session.rs:9` | Artwork URL hand-escaped into JS — only quotes escaped, not backticks/backslashes |
| 7 | `nip66.rs:99` | Unparseable NIP-66 kind maps to 0 (metadata) — misleading relay requirements |
| 8 | `relay_explorer.rs:6` | `std::time::Duration` may panic on WASM targets |
| 9 | `engagement_fetch.rs:51` | `Duration::from_secs` passed to nostr-sdk on potentially WASM targets |
| 10 | `engagement_fetch.rs:118` | Fragile BOLT11 invoice parsing — first `'1'` char heuristic easily tricked |
| 11 | `query_parser.rs:250` | `Timestamp::now()` may panic on WASM when parsing `since:` queries |
| 12 | `group_store/fetch.rs:173` | Membership filter uses `.author(pubkey)` but NIP-28 admin events are authored by admins, not the user |
| 13 | `chat.rs:273` | Edit content `use_effect` may overwrite user input already in progress |
| 14 | `chat.rs:326` | Duplicate ~25-line send logic between Enter keypress and Send button handlers |
| 15 | `use_group_subscription.rs:44` | Fallback listener iterates ALL relay pool notifications — wasteful on busy clients |

---

**💡 Suggestions:**

| # | File | Issue |
|---|------|-------|
| 1 | `publish.rs:49` | All 4 publish functions discard errors with `let _ =` — user sees success for unpublished events |
| 2 | `settings.rs:312` | Delete Group button has no confirmation — accidental irreversible deletion |
| 3 | `share.rs:66` | Web Share API fallback is a no-op — nothing happens for unsupported browsers |
| 4 | `weather_store.rs:12` | Unbounded `WEATHER_DATA` HashMap — no eviction for old locations |
| 5 | `online_status.rs:15` | Polls `document::eval` every 5s — use browser `online`/`offline` events instead |
| 6 | `feed_cache.rs:82` | `Shorts→Verts` rename orphans existing IndexedDB cache entries |
| 7 | `group_store/fetch.rs:228` | Sequential group fetching — use `join_all` for concurrent fetches |
| 8 | `music/player/mod.rs:112` | MPRIS polls every 100ms even when idle — unnecessary CPU wakeups |

---

**✅ What Looks Good:**
- Clean separation of concerns in the viewer architecture (37 viewers dispatched from a single naddr route)
- Proper `cfg` feature gating in most new code
- Correct use of Dioxus hooks and signals throughout
- Good error handling patterns in weather services
- Chess state reconstruction with FEN desync recovery is well-architected
- No security issues found in user content rendering (ammonia sanitization)

---
*Reviewed by [Hermes Agent](https://github.com/apps/hermes-agent)* 🤖

PR #330, opened 2026-05-29
Release 0.8.16: Places with BTCmap, V4V multi-recipient boosts, Cashu IndexedDB persistence

patrickulrich:
Features
- Places — Interactive map with geohash search and BTCmap integration (26 new files, +3601 lines)
- V4V Boosts — Multi-recipient lightning boosts with keysend support
- Cashu Wallet — Upgrade CDK to WalletRepository API with IndexedDB persistence for wallet state

Fixes
- Gap-aware pagination cursors and future-dated event filtering
- Improved thread reliability and repost publishing
- Profile cache chain used for embedded note author lookup
- Simplified note card modals, memoized rich content, reduced feed page size


goodmorningbitcoin:
## 🔍 Code Review — Release 0.8.16

**Verdict: Changes Requested** — 5 critical, 18 warnings, 15 suggestions

This is a substantial PR (88 files, +6,419/-1,058 lines) adding three major features: Places (interactive map + BTCmap), V4V multi-recipient boosts, and Cashu IndexedDB persistence. Reviewed across all three domains.

---

### 🔴 Critical (5)

**1. XSS via `javascript:` URLs in map popups** — `src/components/places/map_container.rs`
OSM `website` field is interpolated directly into `<a href>` in Leaflet popup HTML. A malicious OSM node with `website: "javascript:alert(document.cookie)"` executes arbitrary JS on click. Escape the URL and validate scheme (`http:` or `https:` only) before rendering.

**2. Swapped lat/lng in directions URL** — `src/components/viewers/place_viewer.rs`
`Place.coordinates` follows GeoJSON order `[longitude, latitude]`, but the OpenStreetMap routing URL uses `coordinates[0], coordinates[1]` which puts longitude first. Most routing APIs expect `lat,lon`. The `haversine_km` calls elsewhere correctly use `coordinates[1], coordinates[0]` (lat, lon).

**3. MPP melt no longer atomic** — `src/stores/cashu/mpp.rs`
The old `multi_wallet.mpp_melt(quotes)` was replaced with a sequential loop of `prepare_melt` + `confirm` per mint. If mint #2's melt fails after mint #1 succeeds, proofs are spent but the invoice isn't fully paid. No rollback mechanism exists. Consider wrapping in a compensating transaction pattern or reverting to CDK's atomic multi-mint melt.

**4. `get_incomplete_sagas()` returns ALL sagas** — `src/stores/indexeddb_database.rs`
Returns all sagas from the store without filtering by completion status. On every startup, `recover_cdk_sagas()` will attempt recovery on completed sagas too, causing unnecessary network calls and potential state corruption.

**5. `update_saga` allows same-version overwrite** — `src/stores/indexeddb_database.rs`
The check `saga.version != old.version + 1 && saga.version != old.version` rejects version gaps but **allows same-version overwrites** (`saga.version == old.version`). Two concurrent operations could both pass the check and silently clobber each other's state, defeating optimistic concurrency control entirely. Fix: change to `saga.version != old.version + 1`.

---

### ⚠️ Warnings (18)

**Places (8):**
- **Incomplete JSON escaping in marker data** — `map_container.rs` `format!()` only escapes `\` and `"` but not newlines/tabs, which break the JSON template literal
- **No debounce on geocoding search** — Every keystroke triggers a Nominatim API request (their ToS requires max 1 req/sec). Add 300ms debounce
- **`std::sync::Mutex` for OSM_CACHE in WASM** — In single-threaded WASM, `Mutex::lock()` can poison on panic. Use `RefCell` or `futures::lock::Mutex`
- **Unbounded `PLACES` GlobalSignal** — Vec grows without limit as geohashes are fetched. Add a cap or LRU eviction
- **Share toast never dismisses** — `show_share_toast` set to `true` but never reset to `false`
- **Filter limit of 5000 is excessive** — Dense urban areas at precision 3-4 geohash could return thousands of places, causing rendering lag
- **Deletion events not handled** — `Place.deleted` hardcoded to `false`, no NIP-09 deletion event processing
- **Mutex callback pattern fragile** — `Arc<Mutex<Box<dyn FnMut>>>` can panic on WASM if callback panics while holding the lock

**Cashu (7):**
- **`kv_write` fails on non-UTF-8 binary data** — CDK trait accepts `&[u8]` but implementation calls `String::from_utf8()`. Use base64 encoding
- **`reserve_melt_quote`/`reserve_mint_quote` share the same store** — No type discrimination, same `operation_id` could collide
- **`persist_tokens` silently discards errors** — `let _ = storage::set(...)` means token is in memory but not persisted; funds lost on reload
- **Race condition in `redeem_offline_tokens`** — Concurrent invocations (manual + watcher) can double-redeem the same token
- **Infinite polling with no cancellation** — `start_offline_redemption_watcher` runs every 5s even when no pending tokens exist
- **Full table scans for proofs** — `get_proofs_by_ys`, `get_unissued_mint_quotes`, `get_reserved_proofs` all do O(n) scans
- **`preview_melt_fees` may leave proofs reserved** — If `cancel()` fails after `prepare_melt()`, proofs stay reserved

**V4V/Core (3):**
- **Live interaction count fetching removed** — `note_card.rs` no longer fetches reply/repost/zap counts for notes without precomputed data. Cards outside feeds show 0 for everything
- **Unbounded `NOTIFICATION_EVENT_CACHE`** — `HashMap<String, Event>` grows indefinitely with no eviction policy
- **BFS reply fetch scaling** — `max_rounds=3` × `limit=200` × 3 concurrent queries = up to 1,800 relay requests per thread view

---

### 💡 Suggestions (15)

- **Places**: Opening hours parser handles only a limited subset of OSM format — sunrise/sunset, month ranges, and 24/7 shorthand not supported
- **Places**: Two separate hardcoded amenity category lists (`AMENITY_OPTIONS` in map_container.rs, `AMENITY_CATEGORIES` in places.rs) — consolidate into one source of truth
- **Places**: No input validation on user-submitted place data before publishing (website URL, phone number format)
- **Places**: Hardcoded `email=contact@nostr.blue` in Nominatim API requests — verify this is registered with Nominatim
- **Cashu**: `supports_p2bk()` incorrectly delegates to `supports_p2pk()` — NUT-28 is distinct from NUT-11
- **Cashu**: `is_p2pk_unlock_error` uses fragile string matching — prefer exhaustive CDK error variant matching
- **Cashu**: `sync_wallet_state` called after every operation does full proof scan — consider incremental updates
- **Cashu**: Offline tokens stored unencrypted in localStorage — encrypt with wallet seed
- **Cashu**: `Conditions::new().ok()` silently discards errors in P2PK send — log a warning at minimum
- **V4V**: `blip10.rs` manually builds JSON map — use `#[derive(Serialize)]` for type safety
- **V4V**: Multi-recipient payments are sequential — parallelize with `futures::join_all`
- **V4V**: `total_split` computed but never used (assigned to `_total_split`) — dead code
- **Core**: Publish queue relay selection fallback broadcasts to all write relays instead of gossip — behavior change
- **Core**: `NoteModal` enum defined inside component body — move to module scope
- **Core**: Gap threshold hardcoded to 6 hours in pagination — consider making it configurable

---

### ✅ Looks Good

- Clean separation between Places service, store, and UI layers
- Good use of Dioxus signals for reactive map state
- BFS-based thread fetching is a solid improvement over the previous approach
- Notification event cache avoids redundant relay fetches for the same event
- Cashu migration to WalletRepository API is well-structured overall
- Pagination gap-aware cursors are a smart approach for Nostr event timelines
- Profile cache chain used for embedded note author lookup — reduces network requests
- Good consolidation of 4 boolean signals into `NoteModal` enum in note_card.rs

---

*Reviewed by Hermes Agent*

src/stores/cashu/mpp.rs
```diff
@@ -233,31 +226,37 @@ pub async fn execute_mpp_melt(
             Ok((mint_url, c.quote_id.clone()))
         })
         .collect::<Result<Vec<_>, String>>()?;
-    let results = multi_wallet
-        .mpp_melt(quotes)
-        .await
-        .map_err(|e| format!("MPP melt failed: {}", e))?;
+    let mut melt_results: Vec<(MintUrl, cdk::types::FinalizedMelt)> = Vec::new();
```
goodmorningbitcoin, line 233: 🔴 **Critical: MPP melt no longer atomic**

The old `multi_wallet.mpp_melt(quotes)` delegated atomic multi-mint melting to CDK. The new sequential loop calls `prepare_melt` + `confirm` per mint. If mint #2's melt fails after mint #1 succeeds, proofs are spent but the invoice isn't fully paid — no rollback mechanism exists.

**Options:**
1. Revert to CDK's `mpp_melt()` for atomic multi-mint operations
2. Implement compensating transactions (re-mint spent proofs on failure)
3. At minimum, document this as a known risk and add monitoring for partial payments

src/stores/indexeddb_database.rs
```diff
@@ -1493,10 +1588,132 @@ impl WalletDatabase for IndexedDbDatabase {
...
+                Ok(true)
+            }
+            None => Ok(false),
+        }
+    }
+    async fn delete_saga(&self, id: &uuid::Uuid) -> Result<(), database::Error> {
+        let key = id.to_string();
+        self.delete_value(STORE_SAGAS, &key).await
+    }
+    async fn get_incomplete_sagas(&self) -> Result<Vec<WalletSaga>, database::Error> {
```
goodmorningbitcoin, line 1654: 🔴 **Critical: `get_incomplete_sagas()` returns ALL sagas**

This returns all sagas from the store without filtering for incomplete ones:
```rust
let all: Vec<WalletSaga> = self.get_all_values(STORE_SAGAS).await?;
Ok(all)
```

On every startup, `recover_cdk_sagas()` will attempt recovery on completed sagas too, causing unnecessary network calls and potential state corruption.

**Fix:** Filter by completion status before returning.

src/stores/indexeddb_database.rs
```diff
@@ -1493,10 +1588,132 @@ impl WalletDatabase for IndexedDbDatabase {
...
+    }
+    async fn add_saga(&self, saga: WalletSaga) -> Result<(), database::Error> {
+        let key = saga.id.to_string();
+        self.put_value(STORE_SAGAS, &key, &saga).await
+    }
+    async fn get_saga(&self, id: &uuid::Uuid) -> Result<Option<WalletSaga>, database::Error> {
+        let key = id.to_string();
+        self.get_value(STORE_SAGAS, &key).await
+    }
+    async fn update_saga(&self, saga: WalletSaga) -> Result<bool, database::Error> {
```
goodmorningbitcoin, line 1612: 🔴 **Critical: `update_saga` allows same-version overwrite**

The version check:
```rust
if saga.version != old.version + 1 && saga.version != old.version {
    return Ok(false);
}
```

This **allows** `saga.version == old.version` (same-version overwrite). Two concurrent operations could both pass the check and silently clobber each other's state, defeating optimistic concurrency control entirely.

**Fix:** Change to strict increment-only:
```rust
if saga.version != old.version + 1 {
    return Ok(false);
}
```

goodmorningbitcoin (changes requested):
## Hermes Agent Review — Part 1 (Modified Files)

Critical findings on mpp.rs and indexeddb_database.rs. See summary comment for full review.

goodmorningbitcoin (commented):
## Hermes Agent Review — Part 2 (Places XSS)

src/components/places/map_container.rs
```diff
@@ -0,0 +1,1550 @@
...
+                        ${{hours ? '<div style="margin-bottom:4px;"><span style="color:' + (isOpen === true ? '#4ade80' : isOpen === false ? '#737373' : '#a3a3a3') + ';font-size:12px;">' + (isOpen === true ? '● Open Now' : isOpen === false ? '● Closed' : '') + '</span>' + (fmtHours ? '<div style="color:#737373;font-size:11px;margin-top:2px;">' + fmtHours + '</div>' : '') + '</div>' : ''}}
+                        ${{addrLine ? '<div style="color:#a3a3a3;font-size:12px;margin-bottom:3px;">📍 ' + addrLine + '</div>' : ''}}
+                        ${{phone ? '<div style="margin-bottom:3px;"><a href="tel:' + phone + '" style="color:#a78bfa;text-decoration:none;font-size:12px;">📞 ' + phone + '</a></div>' : ''}}
+                        ${{website ? '<div style="margin-bottom:3px;"><a href="' + website + '" target="_blank" rel="noopener" style="color:#a78bfa;text-decoration:none;font-size:12px;word-break:break-all;">🌐 ' + website.replace(new RegExp("^https?://"), "") + '</a></div>' : ''}}
```
goodmorningbitcoin: 🔴 **Critical: XSS via `javascript:` URLs**

The OSM `website` field is interpolated directly into `<a href>` without scheme validation. A malicious OSM node with `website: "javascript:alert(document.cookie)"` would execute arbitrary JS when a user clicks the link.

**Fix:** Validate the URL scheme before rendering:
```javascript
const safeWebsite = website && /^https?:\/\//i.test(website) ? website : null;
```
Only render the link if `safeWebsite` is non-null.

src/components/viewers/place_viewer.rs
```diff
@@ -0,0 +1,358 @@
...
+        place.address.as_ref().and_then(|a| a.postcode.clone()),
+    ]
+    .into_iter()
+    .flatten()
+    .collect();
+    let address_str = address_parts.join(", ");
+
+    let directions_url = format!(
+        "https://www.openstreetmap.org/directions?from=&to={},{}",
+        place.coordinates[0], place.coordinates[1]
```
goodmorningbitcoin: 🔴 **Critical: Swapped lat/lng in directions URL**

`Place.coordinates` follows GeoJSON order `[longitude, latitude]`, so `coordinates[0]` is longitude and `coordinates[1]` is latitude. This passes them as `coordinates[0], coordinates[1]` (lon, lat) to the routing URL.

Most OpenStreetMap routing services expect `lat,lon` order. Compare with `haversine_km` calls elsewhere in this PR which correctly use `coordinates[1], coordinates[0]` (lat, lon).

**Fix:** Swap to `place.coordinates[1], place.coordinates[0]`.

goodmorningbitcoin (commented):
Hermes Agent Review — Place viewer directions

src/stores/cashu/offline_receive.rs
```diff
@@ -0,0 +1,109 @@
...
+}
+
+pub async fn redeem_offline_tokens() {
+    let tokens_to_redeem = PENDING_OFFLINE_TOKENS.read().clone();
+    if tokens_to_redeem.is_empty() {
+        return;
+    }
+    log::info!(
+        "Starting offline token redemption for {} tokens",
+        tokens_to_redeem.len()
```
goodmorningbitcoin, line 74: ⚠️ **Warning: `persist_tokens` silently discards errors**

`let _ = storage::set(...)` silently discards the result. If localStorage is full or unavailable, the token is in the in-memory GlobalSignal but NOT persisted. On page reload, the token is lost from both IndexedDB (never received) and localStorage (failed to persist). The user's funds exist only in the now-discarded token string.

**Fix:** Propagate the error or at minimum log a warning and surface to the user.

goodmorningbitcoin (commented):
Hermes Agent Review — Cashu offline receive

src/stores/cashu/capabilities.rs
```diff
@@ -102,6 +106,10 @@ impl MintCapabilities {
     pub fn supports_restore(&self) -> bool {
         self.supports_nut(Nut::Restore)
     }
+    /// Check if P2BK (NUT-28) is supported
+    pub fn supports_p2bk(&self) -> bool {
+        self.supports_p2pk()
```
goodmorningbitcoin, line 115: ⚠️ **Warning: `supports_p2bk()` incorrectly delegates to `supports_p2pk()`**

NUT-28 (P2BK) is a distinct protocol from NUT-11 (P2PK). If a mint supports P2PK but not P2BK, this will incorrectly report P2BK support.

**Fix:** Check `mint_info.nuts.nut28` specifically, or at minimum `self.supports_nut(Nut::P2bk)`.

goodmorningbitcoin (commented):
Hermes Agent Review — capabilities.rs warning

src/stores/cashu/mpp.rs
patrickulrich: Fixed in this PR. Investigated deeply — the old CDK `MultiMintWallet::mpp_melt()` (removed in commit b8091315) was **never atomic** either. It spawned parallel tasks and returned `Err` on first failure with no rollback — same partial-failure risk. Our new sequential loop has identical semantics. Also discovered and fixed a more critical bug: `create_mpp_melt_quotes` was passing `None` for MeltOptions instead of `MeltOptions::new_mpp(amount * 1000)`, meaning each mint quoted for the full invoice amount instead of its partial contribution. That fix makes MPP actually work at the protocol level (NUT-15). Doc comments updated to remove false claims about CDK's `mpp_melt()` and a NOTE comment added at the melt loop documenting the non-atomic risk.

src/stores/indexeddb_database.rs
patrickulrich: Not a bug — this is correct by design. Verified against all three CDK database backends (SQL at cdk-sql-common:1535, Redb at cdk-redb:1099, Supabase at cdk-supabase:1832) — they all return ALL rows from `wallet_sagas` with no filtering. CDK deletes completed sagas via `delete_saga()` on every success path across all 5 saga types (swap, send, receive, issue, melt). If a saga exists in the store, it is incomplete by construction. The IndexedDB implementation matches the CDK contract exactly. No change needed.

src/stores/indexeddb_database.rs
patrickulrich: Fixed in this PR. Two changes made:

1. **Strict version check:** Changed from `!= old.version + 1 && != old.version` to strict `!= old.version + 1`, matching all three CDK backends (SQL uses `WHERE version = :expected_version`, Redb checks `existing.version != expected_version`, Supabase filters `version=eq.{expected_version}`).

2. **Single IndexedDB transaction:** Rewrote `update_saga` to use a single `Readwrite` transaction for both the read and write (following the existing `increment_keyset_counter` pattern at line 1383 of this file). The previous implementation used two separate transactions (`get_value` then `put_value`) which had a TOCTOU race window — between the read transaction committing and the write transaction starting, another async task could modify the same saga. The single-transaction approach eliminates this gap and makes the strict version check safe.

src/components/places/map_container.rs
patrickulrich: Fixed in this PR. Added `const safeWeb = website && /^https?:\/\//i.test(m.website) ? website : null;` and changed both popup branches (nostr popups at line 181 and BTCMap popups at line 197) to use `safeWeb` instead of `website` in the href attribute. Only renders the link if the URL scheme is http or https. Also fixed the same vulnerability in `place_viewer.rs` where the Rust-side `href: "{website}"` now filters through the existing `is_valid_http_url()` utility (same pattern used in `radio_viewer.rs:185`). Verified that Dioxus provides zero href sanitization — the `href` attribute value is passed directly to `set_attribute` with no validation (checked dioxus/packages/html/src/elements.rs and dioxus/packages/web/src/mutations.rs).

src/components/viewers/place_viewer.rs
patrickulrich: Fixed in this PR. Swapped to `place.coordinates[1], place.coordinates[0]` (lat, lon). Verified: `coordinates[0]` = longitude, `coordinates[1]` = latitude per GeoJSON standard (confirmed at places.rs:402-404 where `arr[0]` is parsed as `lng` and `arr[1]` as `lat`). All `haversine_km` calls already correctly use `coordinates[1], coordinates[0]` (lat, lon), and the OSRM routing at map_container.rs:374 correctly uses `lng,lat` for OSRM's format (different convention). Only this OSM directions URL was wrong.

src/stores/cashu/offline_receive.rs
patrickulrich: Fixed in this PR. Three changes:

1. `persist_tokens` now returns `Result<(), String>` instead of `()`, propagating the `storage::set` error directly.
2. `store_offline_token` uses `?` to propagate the error — the caller in `receive_modal.rs:90-108` already handles `Err(e)` with a user-facing error message, so it will now properly surface persistence failures.
3. `remove_offline_token` logs a warning on failure via `log::warn!` since it's called from the fire-and-forget `redeem_offline_tokens` loop where propagation isn't practical.

src/stores/cashu/capabilities.rs
patrickulrich: Acknowledged — added a doc comment explaining the design decision. NUT-28 (P2BK) is explicitly documented in CDK as a **client-side-only** feature: *"Compatible with existing mints (no mint-side changes needed)"* (cdk/crates/cashu/src/nuts/nut28/mod.rs:18). The CDK `Nuts` struct has no `nut28` field, so there's no way to populate P2BK from mint info. The delegation to `supports_p2pk()` is correct because P2BK requires P2PK support (NUT-10 + NUT-11) as a prerequisite. Also note: `supports_p2bk()` is currently dead code — confirmed by grep that it's never called anywhere in the codebase. The P2BK feature is always available in the UI whenever P2PK is enabled, without an explicit capability check.

src/stores/cashu/mpp.rs
goodmorningbitcoin: Good investigation. The key finding here is that `create_mpp_melt_quotes` was passing `None` for `MeltOptions` — that was the *actual* bug causing MPP to quote full amounts per mint instead of partial contributions. The sequential loop having identical non-atomic semantics to the old `mpp_melt()` is acceptable given the documentation. That NUT-15 fix is the real win. ✅ Resolved.

src/stores/indexeddb_database.rs
goodmorningbitcoin: Fair point — I stand corrected. If CDK deletes completed sagas on every success path, then any saga remaining in the store is indeed incomplete by construction. The IndexedDB implementation matches the CDK contract. I should have checked the reference implementations before flagging this. Withdrawing this finding. ✅ Resolved — not a bug.

src/stores/indexeddb_database.rs
goodmorningbitcoin: Two solid fixes: (1) strict version-only check matching CDK backends, and (2) single IndexedDB transaction eliminating the TOCTOU window. The single-transaction approach following the `increment_keyset_counter` pattern is the right call — separate read/write transactions in async WASM are inherently racy. ✅ Resolved.

src/components/places/map_container.rs
goodmorningbitcoin: Good fix. The `safeWeb` regex check in both popup branches and the Rust-side `is_valid_http_url()` filter in place_viewer.rs cover both the JS and Dioxus rendering paths. Appreciate the verification that Dioxus provides zero href sanitization — worth knowing for future UI work. ✅ Resolved.

src/components/viewers/place_viewer.rs
goodmorningbitcoin: Clean fix. The cross-reference to places.rs:402-404 confirming GeoJSON `[lng, lat]` order and the note about OSRM using `lng,lat` vs OSM directions using `lat,lon` is helpful context for anyone reading this code later. ✅ Resolved.

src/stores/cashu/offline_receive.rs
goodmorningbitcoin: Good fix on all three counts. Propagating the error through `store_offline_token` with `?` so the existing error handling in `receive_modal.rs` catches it is the right approach. And `log::warn!` for `remove_offline_token` in the fire-and-forget loop is the pragmatic call. ✅ Resolved.

src/stores/cashu/capabilities.rs
goodmorningbitcoin: Acknowledged. If NUT-28 is client-side-only and CDK has no `nut28` field in `Nuts`, there is no mint info to check — the delegation to `supports_p2pk()` as a prerequisite check is correct. The doc comment explaining the reasoning is the right call. Also noted it is dead code currently. Withdrawing this finding. ✅ Resolved — not a bug.

goodmorningbitcoin:
## 🔍 Second-Pass Review — Release 0.8.16 (New Findings)

**Scope:** Quran section (8 new files, 2,449 lines) + fixes from round 1 + second-pass on complex files.

**Previously addressed findings excluded** (MPP melt, update_saga, XSS, lat/lng swap, persist_tokens, supports_p2bk, get_incomplete_sagas, kv_write).

---

### 🔴 Critical (3) — All Quran section

**1. UTF-8 panic in search result truncation** — `src/routes/quran/search.rs:159`
`&result.text[..200.min(result.text.len())]` slices at byte position 200, which will **panic at runtime** when the position falls inside a multi-byte Arabic character or diacritic (tashkeel). Quran text is almost entirely multi-byte UTF-8. Use `text.chars().take(200).collect::<String>()`.

**2. UTF-8 panic in translation chip name** — `src/routes/quran/mod.rs:193`
Same pattern: `&t.english_name[..17.min(t.english_name.len())]` can split a multi-byte character. Edition names from the API can contain non-ASCII characters (transliterated Arabic, non-English languages).

**3. Blocking SQLite I/O in async functions** — `src/services/quran_offline_sqlite.rs`
All `rusqlite::Connection` operations (`execute`, `query_row`, `prepare`) are synchronous and called directly inside `async_trait` methods. On native platforms, this blocks the async runtime, causing UI freezes when reading/writing the full Quran (~5-10MB serialized JSON). Wrap in `tokio::task::spawn_blocking()`. Also: `.expect()` on DB open/pragma/schema will crash the app if the file is corrupted.

---

### ⚠️ Warnings (14)

**Quran (6):**
- **`quran/surah.rs:768`** — `format_selected_ayahs_reference()` has duplicate `surah_name` in format string: `format!("{} {}:{}-{} ({})", surah_name, surah_name, first, last, edition)` produces output like "Al-Baqara Al-Baqara:1-5 (en.asad)". Remove the second `surah_name`.
- **`quran/surah.rs:256,285`** — `selected_ayahs` is a `HashSet<u32>` with arbitrary iteration order. Copy/highlight text may show ayahs in random order. Sort the collected Vec before processing.
- **`quran_offline_indexeddb.rs:88`** — `load_surah()` deserializes the **entire** Quran JSON (~5-10MB) from IndexedDB to extract one surah. Store individual surahs as separate records keyed by `edition:surah_number`.
- **`quran_offline_indexeddb.rs:62`** — `save_complete_quran()` and `delete_edition()` use `let _ = tx.await` which silently ignores IndexedDB transaction errors.
- **`quran_store.rs:156`** — Race condition in `initialize()`: two components mounting simultaneously can both pass the `QURAN_STORE_INITIALIZED` check before either sets it, causing duplicate API calls.
- **`quran_api.rs`** — `fetch_json()` native/reqwest path has no explicit timeout. A hung API call blocks indefinitely.

**Places (5):**
- **`places/map_container.rs`** — Directions button onclick passes `name` through single-quote escaping but doesn't escape backticks or `${`. If the place name contains `` ` `` or `${}`, it can break out of the template literal context.
- **`places/map_container.rs`** — Marker clear-and-rebuild pattern removes ALL markers on every update. For maps with hundreds of markers, this causes visible flicker. No incremental diffing.
- **`services/places.rs`** — `is_place_open()` doesn't handle multi-day ranges like "Mo-Fr". When today is Wednesday and the segment is "Mo-Fr 09:00-17:00", it skips because the segment starts with "Mo". The JS version in map_container.rs correctly handles ranges — Rust/JS inconsistency.
- **`services/places.rs`** — `build_place_event_builder()` formats coordinates as f64 without rounding. Can produce `[0.30000000000000004, ...]`. Round to ~7 decimal places.
- **`services/places.rs`** — `get_unissued_mint_quotes()` filter: if `q.amount` is `None`, `unwrap_or_default()` returns zero, so the `<` comparison is always false, incorrectly classifying quotes with `amount: None` as fully issued.

**Cashu (3):**
- **`indexeddb_database.rs`** — `reserve_melt_quote()` and `reserve_mint_quote()` share `STORE_QUOTE_RESERVATIONS` with no type discriminator. Same `operation_id` can silently overwrite.
- **`indexeddb_database.rs`** — `add_p2pk_key()` uses `std::time::SystemTime::now()` which may panic on WASM. Should use `web_time` or platform timestamp helper.
- **`cashu_cdk_bridge.rs`** — `remove_mint()` passes `CurrencyUnit::Sat` hardcoded. If wallet was created with a different unit, it won't be found for removal.

---

### 💡 Suggestions (12)

- **`quran/surah.rs:29`** — No validation of surah route parameter. `/quran/0` or `/quran/999` will attempt an API call that errors. Add early validation (1..=114).
- **`quran/surah.rs:413`** — Translation ayah matching uses O(n) linear scan per ayah. For Al-Baqara (286 ayahs), this is O(n²) per render. Build a HashMap once.
- **`quran/surah.rs:339`** — Arabic Quran text uses default system fonts. Consider specifying Arabic-optimized font (Amiri, Scheherazade New) for proper diacritic rendering.
- **`quran_store.rs:312`** — `download_edition()` downloads ~5-10MB with no progress reporting or retry logic.
- **`quran_store.rs:101`** — `initialize()` makes 3 independent API calls sequentially. Use `futures::join!` to parallelize.
- **`quran_api.rs:265`** — `get_audio_url()` constructs CDN URLs from reciter identifier without validation. Potential path traversal.
- **`quran_offline_indexeddb.rs:77`** — Entire Quran stored as single JSON string without compression. Consumes IndexedDB quota quickly.
- **`places/map_container.rs`** — No deduplication between nostr places and BTCMap places. Duplicate markers at same location possible.
- **`places/map_container.rs`** — Marker hash uses only data lengths and viewport, not content. Different datasets with same lengths produce same hash.
- **`services/places.rs`** — `geohash_prefix()` silently returns empty string on encode failure. Would match all places in subscription.
- **`cashu/recovery.rs`** — `consolidate_proofs()` swaps all unspent proofs with `SplitTarget::default()`, losing denomination diversity.
- **`cashu_cdk_bridge.rs`** — `init_multi_wallet()` sets `*MULTI_WALLET.write() = None` before building new wallet. If build fails, previous wallet is lost.

---

### ✅ Verified Fixes from Round 1
- XSS in map popups — ✅ Fixed (safeWeb regex + is_valid_http_url)
- Lat/lng swap in directions URL — ✅ Fixed (coordinates[1], coordinates[0])
- MPP melt sequential loop — ✅ Acceptable (CDK's mpp_melt was also non-atomic, NUT-15 fix is the real win)
- update_saga version check — ✅ Fixed (strict increment + single IDB transaction)
- persist_tokens error handling — ✅ Fixed (returns Result, propagated with ?)
- get_incomplete_sagas — ✅ Correct by design (CDK deletes completed sagas)
- kv_write UTF-8 — ✅ Noted (acceptable for current CDK usage)
- supports_p2bk — ✅ Correct (NUT-28 is client-side-only)

---

*Second-pass review by Hermes Agent*

src/routes/quran/search.rs
```diff
@@ -0,0 +1,208 @@
...
+                }
+            }
+        }
+    }
+}
+
+#[component]
+fn SearchResultCard(result: QuranSearchResult, query: String) -> Element {
+    let display_text = if result.text.len() > 200 {
+        format!("{}...", &result.text[..200.min(result.text.len())])
```
goodmorningbitcoin: 🔴 **Critical: UTF-8 panic on Arabic text truncation**

`&result.text[..200.min(result.text.len())]` slices at **byte** position 200, which will **panic at runtime** when that position falls inside a multi-byte Arabic character or diacritic (tashkeel). Quran search results are almost entirely multi-byte UTF-8.

**Fix:** Use char-boundary-aware truncation:
```rust
let display_text: String = result.text.chars().take(200).collect();
let display_text = if result.text.len() > display_text.len() {
    format!("{}...", display_text)
} else {
    display_text
};
```

src/routes/quran/mod.rs
```diff
@@ -0,0 +1,262 @@
...
+    is_favorite: bool,
+    is_downloaded: bool,
+) -> Element {
+    let current_selection = selected.read().clone();
+    let is_selected = current_selection == t.identifier;
+    let tid = t.identifier.clone();
+    let display_name = if t.english_name.len() <= 20 {
+        t.english_name.clone()
+    } else {
+        format!("{}...", &t.english_name[..17.min(t.english_name.len())])
```
goodmorningbitcoin: 🔴 **Critical: UTF-8 panic on translation chip name truncation**

Same pattern as search.rs: `&t.english_name[..17.min(t.english_name.len())]` slices at byte 17, which can split a multi-byte UTF-8 character. Edition names from the API can contain non-ASCII characters (transliterated Arabic, non-English language names).

**Fix:**
```rust
let display_name: String = t.english_name.chars().take(17).collect();
let chip_text = if t.english_name.len() > display_name.len() {
    format!("{}...", display_name)
} else {
    display_name
};
```

src/services/quran_offline_sqlite.rs
```diff
@@ -0,0 +1,84 @@
...
+";
+
+pub struct SqliteQuranStorage {
+    conn: rusqlite::Connection,
+}
+
+impl SqliteQuranStorage {
+    pub fn new() -> Self {
+        let db_path = crate::platform::storage::data_dir()
+            .join("nostr-blue")
```
goodmorningbitcoin, line 18: 🔴 **Critical: Blocking SQLite I/O in async functions**

All `rusqlite::Connection` operations (`execute`, `query_row`) are synchronous and called directly inside `async_trait` methods. On native platforms, this **blocks the async runtime**, causing UI freezes when reading/writing the full Quran (~5-10MB serialized JSON).

**Fix:** Wrap blocking calls in `tokio::task::spawn_blocking()`:
```rust
async fn save_complete_quran(&self, edition_id: &str, data: &CompleteQuranData) -> Result<(), String> {
    let json = serde_json::to_string(data).map_err(|e| format!("Failed to serialize: {}", e))?;
    let conn = self.conn.clone(); // or use Arc<Mutex<Connection>>
    tokio::task::spawn_blocking(move || {
        conn.execute("INSERT OR REPLACE ...", rusqlite::params![edition_id, json])
    }).await.map_err(|e| format!("Spawn error: {}", e))?
    .map_err(|e| format!("DB error: {}", e))
}
```

**Also:** The three `.expect()` calls (DB open, pragmas, schema) will crash the app if the file is corrupted or permissions fail. Consider returning `Result<Self, String>` and handling gracefully.

goodmorningbitcoin (changes requested):
## 🔍 Second-Pass Code Review

Reviewing Quran section + fixes from round 1.

src/routes/quran/surah.rs
```diff
@@ -0,0 +1,773 @@
...
+    surah_name: &str,
+    ayah_nums: &[u32],
+    edition: &str,
+) -> String {
+    if ayah_nums.len() == 1 {
+        format_ayah_reference(surah_name, ayah_nums[0], edition)
+    } else if let (Some(&first), Some(&last)) = (ayah_nums.first(), ayah_nums.last()) {
+        format!(
+            "{} {}:{}-{} ({})",
+            surah_name, surah_name, first, last, edition
```
goodmorningbitcoin: ⚠️ **Warning: Duplicate surah_name in format string**

`format!("{} {}:{}-{} ({})", surah_name, surah_name, first, last, edition)` produces output like:
```
Al-Baqara Al-Baqara:1-5 (en.asad)
```

The second `surah_name` should be removed:
```rust
format!("{}:{}-{} ({})", surah_name, first, last, edition)
```
or if you want the surah name before the range:
```rust
format!("{} {}:{}-{} ({})", surah_name, first, last, edition)
```

src/routes/quran/surah.rs
```diff
@@ -0,0 +1,773 @@
...
+
+    let copy_ayahs = {
+        let selected_ayahs_for_copy = selected_ayahs;
+        let mut selected_ayahs_for_clear = selected_ayahs;
+        let mut show_toolbar_for_clear = show_toolbar;
+        let surah_name_for_copy = surah_name_for_copy.clone();
+        let edition_for_copy = edition_for_copy.clone();
+        move |_| {
+            if let Some(data) = surah_data.read().as_ref() {
+                let selected: Vec<u32> = selected_ayahs_for_copy.read().iter().copied().collect();
```
goodmorningbitcoin: ⚠️ **Warning: Unordered ayah selection**

`selected_ayahs.read().iter().copied().collect()` collects from a `HashSet<u32>` which has **arbitrary iteration order**. When building copy/highlight text, ayahs may appear in random order (e.g., 5, 2, 8, 1) rather than surah order (1, 2, 5, 8).

**Fix:** Sort after collecting:
```rust
let mut selected: Vec<u32> = selected_ayahs_for_copy.read().iter().copied().collect();
selected.sort();
```

src/services/quran_offline_indexeddb.rs
```diff
@@ -0,0 +1,116 @@
...
+            .put_key_val(&js_key, &js_val)
+            .map_err(|e| format!("Failed to put quran: {:?}", e))?;
+        let _ = tx.await;
+        Ok(())
+    }
+
+    async fn load_surah(&self, edition: &str, surah: u32) -> Option<SurahData> {
+        let db = self.open_db().await.ok()?;
+        let tx = db
+            .transaction_on_one_with_mode(STORE, IdbTransactionMode::Readonly)
```
goodmorningbitcoin, line 64: ⚠️ **Warning: Silently ignored IndexedDB transaction errors**

`let _ = tx.await;` discards the transaction result. If the IndexedDB transaction fails (storage quota exceeded, write conflict), the function returns `Ok(())` anyway, giving the user false confidence that their offline Quran was saved.

**Fix:**
```rust
let _ = tx.await.map_err(|e| {
    log::warn!("IndexedDB transaction failed: {:?}", e);
    e
});
```
Or better, propagate the error.

src/stores/quran_store.rs
```diff
@@ -0,0 +1,461 @@
...
+}
+
+pub async fn initialize() -> StdResult<(), String> {
+    if *QURAN_STORE_INITIALIZED.read() {
+        return Ok(());
+    }
+    *LOADING_EDITIONS.write() = true;
+    let result = async {
+        let surahs = fetch_surah_list().await?;
+        *SURAH_LIST.write() = surahs;
```
goodmorningbitcoin, line 109: ⚠️ **Warning: Sequential independent API calls in initialize()**

`fetch_surah_list`, `fetch_editions`, and `fetch_audio_editions` are independent but called sequentially. On slow connections, this triples the initialization time.

**Fix:** Parallelize:
```rust
let (surahs, editions, audio) = futures::try_join!(api.fetch_surah_list(), api.fetch_editions(), api.fetch_audio_editions())?;
```

src/routes/quran/surah.rs
```diff
@@ -0,0 +1,773 @@
...
+    ayah_in_surah: u32,
+) -> MusicTrack {
+    let url = get_audio_url(reciter, ayah_number);
+    MusicTrack {
+        id: format!("quran-{}-{}-{}", surah, reciter, ayah_number),
+        title: format!("{} {}:{}", surah_name, surah, ayah_in_surah),
+        artist: reciter.to_string(),
+        album: Some(format!("Surah {}", surah_name)),
+        media_url: url,
+        album_art_url: None,
```
goodmorningbitcoin, line 29: 💡 **Suggestion: No surah parameter validation**

The route parameter is used directly without bounds checking. Navigating to `/quran/0` or `/quran/999` will attempt an API call that returns an error. Add early validation:
```rust
let surah_num: u32 = surah.parse().unwrap_or(0);
if !(1..=114).contains(&surah_num) {
    return rsx! { "Invalid surah number" };
}
```

src/routes/quran/surah.rs
```diff
@@ -0,0 +1,773 @@
...
+                        div { class: "flex gap-1 items-center",
+                            QuranAudioMenu {
+                                surah,
+                                surah_name: surah_name.clone(),
+                            }
+                            if surah > 1 {
+                                Link {
+                                    to: Route::QuranSurah { surah: surah - 1 },
+                                    class: "p-2 hover:bg-muted rounded-lg transition",
+                                    svg {
```
goodmorningbitcoin, line 356: 💡 **Suggestion: Arabic font for Quran text**

Default system fonts may not render Arabic diacritics (tashkeel/harakat) well, which are essential for Quran recitation accuracy. Consider specifying an Arabic-optimized web font:
```css
font-family: 'Amiri', 'Scheherazade New', 'Noto Naskh Arabic', serif;
```
This could be loaded via a Google Fonts link or bundled as a static asset.

goodmorningbitcoin (commented):
Quran section warnings.

goodmorningbitcoin (commented):
Places and Cashu warnings from second pass.

src/components/places/map_container.rs
```diff
@@ -0,0 +1,1551 @@
...
+                        ${{addrLine ? '<div style="color:#a3a3a3;font-size:12px;margin-bottom:3px;">📍 ' + addrLine + '</div>' : ''}}
+                        ${{phone ? '<div style="margin-bottom:3px;"><a href="tel:' + phone + '" style="color:#a78bfa;text-decoration:none;font-size:12px;">📞 ' + phone + '</a></div>' : ''}}
+                        ${{safeWeb ? '<div style="margin-bottom:3px;"><a href="' + safeWeb + '" target="_blank" rel="noopener" style="color:#a78bfa;text-decoration:none;font-size:12px;word-break:break-all;">🌐 ' + safeWeb.replace(new RegExp("^https?://"), "") + '</a></div>' : ''}}
+                        <div style="display:flex;gap:6px;margin-top:8px;">
+                            ${{s(m.naddr) ? '<a href="/' + s(m.naddr) + '" style="padding:5px 14px;border-radius:6px;background:transparent;color:#a78bfa;border:1px solid #7c3aed;cursor:pointer;font-size:12px;font-weight:500;text-decoration:none;">View Details</a>' : ''}}
+                            <button onclick="window.__placesRequestDirections(${{m.lat}},${{m.lng}},'${{name.replace(/'/g, "\\\\'")}}')"
```
goodmorningbitcoin, line 185: ⚠️ **Warning: Potential injection in directions onclick**

The directions button passes `name` through single-quote escaping but doesn't escape backticks or template literal syntax (`${}`). If a place name contains a backtick or `${...}`, it can break out of the template literal context.

The XSS fix from round 1 correctly handles the `website` link, but this onclick handler uses raw `m.name`.

**Fix:** Apply `esc()` to the name in the onclick, or use a data attribute + event delegation.

src/components/places/map_container.rs
```diff
@@ -0,0 +1,1551 @@
...
+
+        spawn(async move {
+            let clear_js = format!(
+                r#"
+                (() => {{
+                    const maps = window.leafletMaps || new Map();
+                    const map = maps.get({id_json});
+                    if (!map) return;
+                    map.eachLayer(layer => {{
+                        if (layer instanceof L.Marker && layer !== window.__placesUserDot) map.removeLayer(layer);
```
goodmorningbitcoin, line 857: ⚠️ **Warning: Marker clear-and-rebuild causes flicker**

This removes ALL markers on every update, then recreates them. For dense urban areas with hundreds of markers, this causes visible flicker.

Consider tracking existing markers by ID and only adding/removing deltas.

src/stores/indexeddb_database.rs
```diff
@@ -1493,10 +1588,156 @@ impl WalletDatabase for IndexedDbDatabase {
...
+        let val_str = String::from_utf8(value.to_vec())
+            .map_err(|e| Self::make_error(format!("KV value not valid UTF-8: {}", e)))?;
+        self.put_value(STORE_KV, &composite_key, &val_str).await
+    }
+    async fn kv_remove(&self, primary_namespace: &str, secondary_namespace: &str, key: &str) -> Result<(), database::Error> {
+        let composite_key = format!("{}:{}:{}", primary_namespace, secondary_namespace, key);
+        self.delete_value(STORE_KV, &composite_key).await
+    }
+    async fn add_p2pk_key(&self, pubkey: &CashuPublicKey, derivation_path: DerivationPath, derivation_index: u32) -> Result<(), database::Error> {
+        let now = std::time::SystemTime::now()
```
goodmorningbitcoin: ⚠️ **Warning: `std::time::SystemTime` in `add_p2pk_key` may panic on WASM**

`std::time::SystemTime::now().duration_since(std::time::UNIX_EPOCH)` may panic on WASM targets. The codebase already has `web_time` and `crate::platform::timestamp::now_secs()`.

**Fix:** `let now = crate::platform::timestamp::now_secs() as u64;`

src/stores/cashu_cdk_bridge.rs
```diff
@@ -29,112 +31,122 @@ pub struct WalletBalances {
...
         .read()
         .as_ref()
-        .ok_or("MultiMintWallet not initialized")?
+        .ok_or("WalletRepository not initialized")?
         .clone();
     let mint_url = mint_url
         .parse()
         .map_err(|e| format!("Invalid mint URL: {}", e))?;
-    multi_wallet.remove_mint(&mint_url).await;
+    repo.remove_wallet(mint_url, CurrencyUnit::Sat)
```
goodmorningbitcoin, line 82: ⚠️ **Warning: `remove_mint()` hardcodes `CurrencyUnit::Sat`**

If a wallet was added with a different currency unit (e.g., `msat`), it won't be found for removal.

Consider tracking the currency unit alongside the mint URL, or iterating all wallets for that mint.

PR #333, opened 2026-06-02
feat: topics NIP-73 migration, followers modal, relay-aware live chat, and 30+ bug fixes

patrickulrich:
- **Topics → NIP-73**: Migrate topic communities from hashtags to web URL namespaces with DB-first loading, real-time subscriptions, and live update banners
- **Followers/Following modal**: New `FollowersModal` component with pagination, profile resolution, follow/unfollow actions, backed by nostrarchives social graph API
- **Relay-aware live chat**: Refactor `LiveChat` to connect ephemeral relays for stream-specific chat messages

## Fixes
- **Nests audio**: Rewrite moq-nest.js with `@moq/net` library, fix WebTransport URL/auth, fix heartbeat loop
- **Feed stability**: Gap-aware pagination to prevent outlier cursor jumps; always apply `since` filter for partial caches
- **Notification busy-loop**: Replace `loop { if let Ok }` with `while let Ok` in dispatcher
- **Subscription leaks**: Add `AtomicBool` cancellation guards to relay/group subscription hooks
- **Memory safety**: LruCache for interaction store, stale profile eviction, wallet seed zeroization
- **Mutex poison resilience**: Recover from poisoned mutexes across 10 files instead of panicking
- **Bolt11 parsing**: Replace hand-rolled parsers with `lightning-invoice` crate; fix sub-satoshi truncation
- **Blossom validation**: HTTPS-only enforcement, private IP blocking, credential injection protection
- **Search reactivity**: Memo+effect chain so autocomplete updates on session restore
- **Note threads**: DB-first parent loading, NIP-22 comment support, orphaned node preservation
- **Cashu**: Remove broken DLEQ pre-check, defer draft saves, eliminate double-parses
- **NIP-78 double-fire**: Merge retry effects into single reactive gate
- **Path security**: Reject NUL bytes in `is_safe_path()`
- **Nest reactions**: Guard cleanup loop with `is_joined` check

### DIFF
--- a/public/moq-nest.js
+++ b/public/moq-nest.js
@@ -8,7 +8,7 @@ window.nestAudioManager = window.nestAudioManager || {
         if (this._moqLoading) return this._moqLoading;
         this._moqLoading = (async () => {
             try {
-                this._moqModule = await import('https://esm.sh/@kixelated/moq@0.9.4');
+                this._moqModule = await import('https://esm.sh/@moq/net@0.1.2');
                 return this._moqModule;
             } catch (e) {
                 this._moqLoading = null;
@@ -26,15 +26,14 @@ window.nestAudioManager = window.nestAudioManager || {
         if (this.connections.has(publisherId)) return { type: 'success' };
         this.connections.set(publisherId, {
             state: 'disconnected',
-            session: null,
-            publisher: null,
+            connection: null,
+            myPubkeyHex: null,
+            publisherBroadcast: null,
+            publisherTracks: new Set(),
             subscribers: new Map(),
             audioEncoder: null,
-            audioDecoder: null,
-            audioContext: null,
             mediaStream: null,
             micMuted: false,
-            namespace: null,
             error: null,
             participantTracks: [],
         });
@@ -51,15 +50,17 @@ window.nestAudioManager = window.nestAudioManager || {
         }
     },
 
-    async connect(publisherId, authUrl, relayUrl, namespace, jwt) {
+    async connect(publisherId, authUrl, relayUrl, namespace, jwt, myPubkeyHex) {
         const conn = this._getConnection(publisherId);
         if (!conn) return { type: 'error', error: 'Not initialized' };
         try {
             conn.state = 'connecting';
+            const Moq = this._moqModule;
             const url = new URL(relayUrl);
+            url.pathname = '/' + namespace;
             if (jwt) url.searchParams.set('jwt', jwt);
-            conn.session = new this._moqModule.Session(url.toString());
-            conn.namespace = namespace;
+            conn.connection = await Moq.Connection.connect(url);
+            conn.myPubkeyHex = myPubkeyHex;
             conn.state = 'connected';
             return { type: 'success' };
         } catch (e) {
@@ -72,21 +73,51 @@ window.nestAudioManager = window.nestAudioManager || {
     async publishAudio(publisherId) {
         const conn = this._getConnection(publisherId);
         if (!conn) return { type: 'error', error: 'Not initialized' };
-        if (!conn.session) return { type: 'error', error: 'Not connected' };
+        if (!conn.connection) return { type: 'error', error: 'Not connected' };
+        if (conn.publisherBroadcast) return { type: 'success' };
         try {
+            const Moq = this._moqModule;
             conn.state = 'publishing';
+
+            const broadcast = new Moq.Broadcast();
+            conn.publisherBroadcast = broadcast;
+            conn.connection.publish(Moq.Path.from(conn.myPubkeyHex), broadcast);
+
+            (async () => {
+                try {
+                    for (;;) {
+                        const request = await broadcast.requested();
+                        if (!request) break;
+                        if (request.track.name === 'audio/data') {
+                            conn.publisherTracks.add(request.track);
+                            request.track.closed.then(() => {
+                                conn.publisherTracks.delete(request.track);
+                            });
+                        }
+                    }
+                } catch (e) {
+                    if (conn.state === 'publishing') {
+                        console.error('[MoQ Nest] Publisher requested() error:', e);
+                    }
+                }
+            })();
+
             conn.mediaStream = await navigator.mediaDevices.getUserMedia({ audio: true });
-            conn.audioContext = new AudioContext({ sampleRate: 48000 });
-            const source = conn.audioContext.createMediaStreamSource(conn.mediaStream);
             const processor = new MediaStreamTrackProcessor({ track: conn.mediaStream.getAudioTracks()[0] });
             const reader = processor.readable.getReader();
-            conn.publisher = conn.session.publish(conn.namespace);
-            const track = conn.publisher.createTrack('audio');
+
             conn.audioEncoder = new AudioEncoder({
                 output: (chunk, _metadata) => {
                     const buffer = new ArrayBuffer(chunk.byteLength);
                     chunk.copyTo(buffer);
-                    track.writeChunk(buffer);
+                    const data = new Uint8Array(buffer);
+                    const varintBytes = Moq.Varint.encode(chunk.timestamp);
+                    const frame = new Uint8Array(varintBytes.length + data.length);
+                    frame.set(varintBytes, 0);
+                    frame.set(data, varintBytes.length);
+                    for (const track of conn.publisherTracks) {
+                        try { track.writeFrame(frame); } catch (_) {}
+                    }
                 },
                 error: (e) => {
                     console.error('[MoQ Nest] AudioEncoder error:', e);
@@ -98,13 +129,13 @@ window.nestAudioManager = window.nestAudioManager || {
                 numberOfChannels: 1,
                 bitrate: 64000,
             });
+
             (async () => {
                 try {
                     while (true) {
                         const result = await reader.read();
                         if (result.done) break;
-                        if (conn.micMuted) continue;
-                        if (conn.audioEncoder.state === 'configured') {
+                        if (!conn.micMuted && conn.audioEncoder.state === 'configured') {
                             conn.audioEncoder.encode(result.value);
                         }
                         result.value.close();
@@ -126,13 +157,16 @@ window.nestAudioManager = window.nestAudioManager || {
     async subscribeAudio(publisherId, participantPubkey) {
         const conn = this._getConnection(publisherId);
         if (!conn) return { type: 'error', error: 'Not initialized' };
-        if (!conn.session) return { type: 'error', error: 'Not connected' };
+        if (!conn.connection) return { type: 'error', error: 'Not connected' };
         if (conn.subscribers.has(participantPubkey)) return { type: 'success' };
         try {
-            const trackName = participantPubkey + '/audio';
-            const subscription = conn.session.subscribe(conn.namespace, trackName);
+            const Moq = this._moqModule;
+            const broadcast = conn.connection.consume(Moq.Path.from(participantPubkey));
+            const track = broadcast.subscribe('audio/data', 128);
+
             const subState = {
-                subscription,
+                broadcast,
+                track,
                 audioContext: new AudioContext({ sampleRate: 48000 }),
                 audioDecoder: null,
                 active: true,
@@ -178,17 +212,17 @@ window.nestAudioManager = window.nestAudioManager || {
                 sampleRate: 48000,
                 numberOfChannels: 1,
             });
-            let frameIndex = 0;
             (async () => {
                 try {
                     while (subState.active) {
-                        const chunk = await subscription.readChunk();
-                        if (!subState.active) break;
+                        const frame = await track.readFrame();
+                        if (!subState.active || !frame) break;
+                        const [timestampUs, payload] = Moq.Varint.decode(frame);
                         if (subState.audioDecoder.state === 'configured') {
                             const data = new EncodedAudioChunk({
                                 type: 'key',
-                                timestamp: frameIndex++ * 20000,
-                                data: chunk,
+                                timestamp: timestampUs,
+                                data: payload,
                             });
                             subState.audioDecoder.decode(data);
                         }
@@ -215,6 +249,7 @@ window.nestAudioManager = window.nestAudioManager || {
         const subState = conn.subscribers.get(participantPubkey);
         if (subState) {
             subState.active = false;
+            try { if (subState.track) subState.track.close(); } catch (_) {}
             try { if (subState.audioDecoder) subState.audioDecoder.close(); } catch (_) {}
             try { if (subState.audioContext) subState.audioContext.close(); } catch (_) {}
             conn.subscribers.delete(participantPubkey);
@@ -246,6 +281,7 @@ window.nestAudioManager = window.nestAudioManager || {
         try {
             for (const [_pubkey, sub] of conn.subscribers.entries()) {
                 sub.active = false;
+                try { if (sub.track) sub.track.close(); } catch (_) {}
                 try { if (sub.audioDecoder) sub.audioDecoder.close(); } catch (_) {}
                 try { if (sub.audioContext) sub.audioContext.close(); } catch (_) {}
             }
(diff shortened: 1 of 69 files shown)


src/routes/topics/home.rs
```diff
@@ -30,44 +35,224 @@ pub fn TopicsHome() -> Element {
...
+                {
+                    Ok(sub_id) => {
+                        subscription_ids.write().push(sub_id.clone());
+                        let active_ids = subscription_ids;
+                        let mut pending = pending_posts;
+                        let current_posts = posts;
+                        let _current_votes = vote_counts;
+                        spawn(async move {
+                            let mut notifications = client.notifications();
+                            while let Ok(notification) = notifications.recv().await {
```
goodmorningbitcoin: 🔴 **Task leak — no cancellation guard**

This `spawn` creates a long-lived async task reading `client.notifications()` in a `while let Ok` loop with no termination mechanism. When the user navigates to a different topic or the component re-renders, a new listener is spawned but the old one keeps running indefinitely.

Over time, navigating between topics accumulates leaked tasks that each consume memory and CPU processing notifications they'll eventually filter out via `subscription_ids`.

**Fix:** Use the `AtomicBool` cancellation pattern already established in `use_relay_subscription` and `use_group_subscription`. Set it in `use_drop` and check it in the `while let` loop:

```rust
let cancel = use_hook(|| Arc::new(AtomicBool::new(false)));
// In spawn:
while !cancel.load(Ordering::Relaxed) && let Ok(notification) = notifications.recv().await {
    // ...
}
```

src/routes/topics/topic_feed.rs
```diff
@@ -28,33 +29,162 @@ pub fn TopicFeed(topic: String) -> Element {
...
+                {
+                    Ok(sub_id) => {
+                        subscription_ids.write().push(sub_id.clone());
+                        let active_ids = subscription_ids;
+                        let mut pending = pending_posts;
+                        let current_posts = posts;
+                        let _current_votes = vote_counts;
+                        spawn(async move {
+                            let mut notifications = client.notifications();
+                            while let Ok(notification) = notifications.recv().await {
```
goodmorningbitcoin: 🔴 **Task leak — same issue as `topics/home.rs:182`**

Identical pattern: `spawn` with `while let Ok(notification) = notifications.recv().await` loop with no cancellation. Every navigation to a new topic leaks the previous listener task.

Apply the same `AtomicBool` fix as suggested for `topics/home.rs`.

public/moq-nest.js
```diff
@@ -51,15 +50,17 @@ window.nestAudioManager = window.nestAudioManager || {
...
-    async connect(publisherId, authUrl, relayUrl, namespace, jwt) {
+    async connect(publisherId, authUrl, relayUrl, namespace, jwt, myPubkeyHex) {
         const conn = this._getConnection(publisherId);
         if (!conn) return { type: 'error', error: 'Not initialized' };
         try {
             conn.state = 'connecting';
+            const Moq = this._moqModule;
             const url = new URL(relayUrl);
+            url.pathname = '/' + namespace;
             if (jwt) url.searchParams.set('jwt', jwt);
```
goodmorningbitcoin, line 66: 🔴 **Security: JWT in URL query parameter**

```js
if (jwt) url.searchParams.set('jwt', jwt);
```

JWTs passed as URL query parameters are:
- Logged by web servers, reverse proxies, and CDNs
- Saved in browser history
- Leaked via `Referer` headers to any cross-origin resources
- Visible in server access logs

**Fix:** If the MoQ protocol supports it, pass the JWT via a header (e.g., `Authorization: Bearer <jwt>`). If URL-based auth is mandatory for WebTransport, ensure the JWT has a very short TTL (e.g., 60 seconds) and document the trade-off.

src/services/nests_audio/mod.rs
```diff
@@ -218,18 +221,35 @@ pub async fn js_get_participant_tracks(publisher_id: &str) -> Vec<String> {
...
+) -> Result<String, String> {
     use crate::platform::http::http_client;
-    use crate::utils::nips::nip98::{create_auth_header, AuthResult};
+    use crate::utils::nips::nip98::{create_auth_header_with_payload, AuthResult};
+    use bitcoin_hashes::Hash as _;
     use nostr_sdk::nips::nip98;
 
+    let full_url = format!("{}/auth", auth_url.trim_end_matches('/'));
+
+    let body = format!(
```
goodmorningbitcoin: ⚠️ **JSON injection via manual string formatting**

```rust
let body = format!(
    "{{\"namespace\":\"{}\",\"publish\":{}}}",
    namespace, publish
);
```

If `namespace` contains `"`, `{`, `}`, or other JSON-special characters, this produces malformed JSON or injects arbitrary fields. While `namespace` is currently a trusted value, this is fragile.

**Fix:** Use `serde_json` for proper serialization:

```rust
let body = serde_json::json!({
    "namespace": namespace,
    "publish": publish
}).to_string();
```

src/utils/validation.rs
```diff
@@ -119,6 +119,82 @@ pub fn sanitize_lightning_invoice(invoice: &str) -> Option<String> {
...
+pub fn validate_blossom_server_url(input: &str) -> Result<String, String> {
+    let trimmed = input.trim();
+    if trimmed.is_empty() {
+        return Err("URL cannot be empty".to_string());
+    }
+    let url_str = if !trimmed.starts_with("https://") {
+        if trimmed.starts_with("http://") {
+            return Err("Blossom servers must use HTTPS".to_string());
+        }
+        format!("https://{}", trimmed)
```
goodmorningbitcoin, line 131: ⚠️ **URL path silently stripped during normalization**

The normalization logic strips the URL path (only scheme + host + port are preserved). A user entering `https://cdn.example.com/blossom` silently gets `https://cdn.example.com` — they'd connect to the wrong endpoint without any indication.

**Fix:** Either preserve the path in the normalized URL, or return an error/warning when a non-empty path would be discarded:

```rust
if url.path() != "/" {
    return Err("Blossom server URLs should not include a path. Remove the path and try again.".to_string());
}
```

src/components/viewers/nest_viewer.rs
```diff
@@ -279,7 +279,7 @@ pub fn NestViewer(naddr: String) -> Element {
                 loop {
                     crate::platform::timer::sleep_ms(60_000).await;
                     if !*is_joined_hb.read() {
```
goodmorningbitcoin, line 281: ⚠️ **`break` → `continue` creates immortal background task**

The heartbeat loop previously used `break` when `!is_joined`, which terminated the spawned task. Now it uses `continue`, so the task loops forever (sleeping 60s per iteration) even after the user leaves the nest.

While the 60s sleep mitigates CPU impact, the task is never cleaned up and will persist for the lifetime of the app session.

**Fix:** Keep `break` to terminate the task when not joined, or use an `AtomicBool` cancellation guard in `use_drop`.

src/utils/parsing/thread_tree.rs
```diff
@@ -338,6 +338,9 @@ pub fn build_thread_tree(replies: Vec<Event>, root_event_id: &EventId) -> Vec<Th
     for node in &mut root_replies {
         node.children = attach_children(&node.event.id, &replies, &mut node_map);
     }
+    for (_, orphan) in node_map.drain() {
```
goodmorningbitcoin: ⚠️ **Orphan nodes appended as root-level replies**

```rust
for (_, orphan) in node_map.drain() {
    root_replies.push(orphan);
}
```

Events whose parent wasn't in the fetched set are appended as top-level replies. This breaks thread visual hierarchy — a reply to a comment 5 levels deep will appear at the root level.

**Suggestion:** Log when orphans are detected so developers can diagnose missing parent events. Optionally, add an `is_orphan: bool` flag to `ThreadNode` so the UI can render them differently (e.g., "replying to [unavailable]").

goodmorningbitcoin (commented):
## 🤖 Hermes Agent Code Review — PR #333

**64 files changed, +2,568 / −787 lines**

Reviewed against `.coderabbit.yml` guidelines for Rust/Dioxus Nostr client targeting WASM/Android/Desktop.

### 🔴 Critical Issues (4)
1. **`followers_modal.rs:53`** — Early return before hooks violates Dioxus Rules of Hooks
2. **`topics/home.rs:182`** — Notification listener task leak (no cancellation guard)
3. **`topic_feed.rs:136`** — Same notification listener task leak
4. **`moq-nest.js:61`** — JWT passed as URL query parameter (security)

### ⚠️ Warnings (7)
5. **`nests_audio/mod.rs:236`** — Manual JSON formatting vulnerable to injection
6. **`followers_modal.rs:78`** — `initial_load_done` never resets on pubkey change
7. **`validation.rs:131`** — URL path silently stripped during normalization
8. **`nest_viewer.rs:281`** — `break→continue` creates immortal background task
9. **`thread_tree.rs:341`** — Orphan nodes appended as root replies
10. **`notification_dispatcher.rs:74`** — Non-Event notification variants silently discarded
11. **`social_graph.rs:9`** — Hard dependency on external API without fallback

### 💡 Suggestions (3)
12. **`nip53.rs`** — Empty `providers` no longer rejected (NIP-53 spec ambiguity)
13. **`topics/home.rs` & `topic_feed.rs`** — Duplicated `accept_pending_posts` logic
14. **`moq-nest.js`** — Dynamic ESM import from CDN without SRI

---
*Reviewed by Hermes Agent — see inline comments for details*

src/stores/notification_dispatcher.rs
```diff
@@ -71,28 +71,27 @@ impl NotificationDispatcher {
 
                 rt.block_on(async move {
                     let mut notifications = client.notifications();
-                    loop {
-                        if let Ok(RelayPoolNotification::Event {
-                            subscription_id,
-                            event,
-                            ..
-                        }) = notifications.recv().await
+                    while let Ok(RelayPoolNotification::Event {
```
goodmorningbitcoin: 💡 **Non-Event notification variants silently discarded**

```rust
while let Ok(RelayPoolNotification::Event {
    subscription_id,
    event,
    ..
}) = notifications.recv().await
```

The `while let Ok(pattern)` destructuring silently filters out `RelayPoolNotification::Shutdown`, `RelayPoolNotification::Message`, and other variants. This means relay disconnections and other lifecycle events are never processed.

This is likely intentional (only care about events), but worth noting that relay pool shutdown won't gracefully terminate the listener — the `recv()` will eventually return `Err` which exits the loop.

Consider adding an explicit match on `Shutdown` to log or break cleanly.

src/components/followers_modal.rs
```diff
@@ -0,0 +1,522 @@
...
+}
+
+fn get_picture_from_metadata(meta: &ProfileMetadata) -> Option<String> {
+    meta.picture.as_ref().filter(|p| !p.trim().is_empty()).cloned()
+}
+
+#[component]
+pub fn FollowersModal(props: FollowersModalProps) -> Element {
+    let mut open = props.open;
+    if !*open.read() {
```
goodmorningbitcoin: 🔴 **Critical: Early return before hooks violates Dioxus Rules of Hooks**

```rust
if !*open.read() {
    return rsx! {};
}
// hooks start at line 57:
let mut active_tab = use_signal(|| props.initial_tab);
```

Dioxus (like React) requires hooks to be called **unconditionally** on every render. The early return on line 53-54 means that if `open` transitions from `true` → `false` → `true`, the component will skip hooks on the false render, corrupting Dioxus's internal hook index and causing panics or undefined behavior.

**Fix:** Move all hooks above the early return, and conditionally render only the modal body:

```rust
// Always call hooks first
let mut active_tab = use_signal(|| props.initial_tab);
let mut following_entries = use_signal(Vec::<UserEntry>::new);
// ... all other hooks ...

// Early return for rendering only
if !*open.read() {
    return rsx! {};
}
// render modal body
```

src/components/followers_modal.rs
```diff
@@ -0,0 +1,522 @@
...
+            .map(|pk| pk.to_hex())
+            .unwrap_or_else(|| props.pubkey.clone())
+    };
+
+    let hex_for_initial = hex_pubkey.clone();
+    let hex_for_following_tab = hex_pubkey.clone();
+    let hex_for_followers_tab = hex_pubkey.clone();
+    let hex_for_load_more = hex_pubkey.clone();
+
+    use_effect(use_reactive(&(*open.read()), move |is_open| {
```
goodmorningbitcoin: ⚠️ **`initial_load_done` never resets on pubkey change**

```rust
let mut initial_load_done = use_signal(|| false);
```

This signal is set to `true` after the first load but never reset. If the modal closes and reopens for a **different user** (different `props.pubkey`), `initial_load_done` is still `true` from the previous user, so the fetch is skipped entirely.

**Fix:** Reset `initial_load_done` when `props.pubkey` changes:

```rust
use_effect(use_reactive(&props.pubkey, move |_| {
    initial_load_done.set(false);
}));
```

Or use a `pubkey` guard: only skip loading if the pubkey matches the one that was loaded.

src/services/social_graph.rs
```diff
@@ -0,0 +1,194 @@
+use serde::{Deserialize, Serialize};
+#[cfg(feature = "web")]
+use wasm_bindgen::JsCast;
+#[cfg(feature = "web")]
+use wasm_bindgen_futures::JsFuture;
+#[cfg(feature = "web")]
+use web_sys::{Request, RequestInit, RequestMode, Response};
+
+const NOSTR_ARCHIVES_API: &str = "https://api.nostrarchives.com";
```
goodmorningbitcoin, line 9: ⚠️ **Hard dependency on external API without fallback**

```rust
const NOSTR_ARCHIVES_API: &str = "https://api.nostrarchives.com";
```

The entire `FollowersModal` depends on `api.nostrarchives.com` for social graph data. If this service is down or the user is behind a firewall that blocks it:
- Followers/following counts show as 0
- No profiles are loaded
- No follow/unfollow actions work

**Suggestions:**
1. Add a timeout and graceful error handling (404 → "Service unavailable")
2. Consider a Nostr-native fallback using contact list events (NIP-02) and relay queries
3. Cache results locally to work offline
4. Add a loading/error state to the modal UI for when the API is unreachable

goodmorningbitcoin (commented):
Review batch 2: inline comments on new files (followers_modal.rs, social_graph.rs)

goodmorningbitcoin:
## 🤖 Hermes Agent — Additional Findings (Lines Outside Diff Hunks / Cross-File Observations)

### 🔴 Supply Chain Risk — `public/moq-nest.js`
The rewritten `moq-nest.js` dynamically imports `@kixelated/moq@0.9.4` from `esm.sh` CDN at runtime:
```js
const { default: Moq } = await import("https://esm.sh/@kixelated/moq@0.9.4");
```
- No Subresource Integrity (SRI) check on the imported module
- If `esm.sh` is compromised or the package is removed, the app breaks silently
- The pinned version `0.9.4` mitigates some risk but not CDN compromise

**Suggestion:** Bundle the MoQ library locally instead of loading from CDN, or use an importmap with SRI hashes. At minimum, add a CSP `script-src` that restricts to `esm.sh` only.

---

### ⚠️ NIP-53 Validation Relaxed — `src/utils/nips/nip53.rs`
The `providers.is_empty()` check was **removed** from `parse_meeting_space()`. A kind 31922 event with zero `p` tags (no participants/hosts) will now be treated as a valid `MeetingSpace`. Downstream code that iterates `providers` should be audited to handle empty vectors gracefully.

---

### 💡 Duplicated Logic — `accept_pending_posts`
The `accept_pending_posts` closure is duplicated verbatim between:
- `src/routes/topics/home.rs` (~line 312)
- `src/routes/topics/topic_feed.rs` (~line 202)

Both implementations: drain `pending_posts` → dedup with `HashSet` → sort by `created_at` → set `posts`. Consider extracting to a shared utility.

---

### 💡 `lightning-invoice` WASM Compatibility
The new `bolt11.rs` uses `lightning-invoice = "0.34.0"` (with `default-features = false`). This pulls in `secp256k1` which may not compile on `wasm32-unknown-unknown` without the `wasm` feature flag. Verify that `cargo build --target wasm32-unknown-unknown --features web` succeeds before merge.

---

### 💡 Bolt11 Unit Consistency
Two functions were introduced:
- `parse_bolt11_amount()` → returns **sats** (via `div_ceil(1000)`)
- `parse_bolt11_amount_msats()` → returns **millisatoshis**

Callers in `zap_goals_store.rs`, `nostr_music.rs`, and `pin_boards_store/fetch.rs` previously had local parsers. Verify each caller is using the correct function (sats vs msats) to avoid double-conversion bugs.

---

### ✅ What Looks Good
- **Mutex poison resilience** — systematic conversion to `unwrap_or_else(|e| e.into_inner())` across 10+ files
- **AtomicBool cancellation guards** in `use_relay_subscription` and `use_group_subscription` — clean pattern
- **DB-first loading** — topics, notes, and profiles load from local DB immediately, then merge relay results
- **Blossom SSRF protection** — comprehensive private IP blocking (IPv4 + IPv6), HTTPS enforcement, credential injection prevention
- **Seed zeroization** — `ZeroizedSeed` wrapper with `Drop` impl that calls `.zeroize()`
- **DLEQ soft-fail** — kept verification but made non-blocking (logs warning, continues receive)

---
*End of review — Hermes Agent*

PR #341, opened 2026-06-26
feat: v0.8.18 — Mostro P2P exchange, Nests rebuild, unified prefs, and feeds infrastructure

patrickulrich:
Bumps version **0.8.17 → 0.8.18**. This release adds the Mostro P2P exchange integration,
rebuilds the Nests live-audio feature, consolidates user preferences into unified NIP-78 blobs,
introduces a new feeds infrastructure, and removes the Blobbi virtual pet subsystem.

**414 files changed, +45,222 / −21,662** across 31 commits.

---

### Mostro P2P exchange integration
Full rebrand of the P2P feature to **Mostro** with end-to-end trade execution:
- **Protocol v2** (`mostro-core` 0.11.5 → 0.13.2): NIP-44 direct (kind 14) transport alongside
  legacy NIP-59 gift-wrap, with transport-aware filters and PoW-first-contact spam gating.
- Complete trade flow: order book (NIP-69 kind 38383), "Take with Mostro", create-order,
  live trade detail (chat, hold-invoice panel, status timeline), dispute resolution, and ratings.
- **Notifications** store, **solver console** (admin disputes/settlements), and **Web Push**.
- NIP-78 terms-of-service gating with signature verification + offline-resilient local cache.
- New `/settings/p2p` page: Mostro mnemonic, privacy mode, daemon discovery, node capabilities.
- Deep-link handler, toast drainer, encrypted attachments, and i18n for trade messages.

### Nests (NIP-53) rebuild
Overhauled to reach feature parity with Amethyst / NestsUI-v2 reference clients:
- Persistent `NEST_ROOM` singleton — audio survives navigation (like the music player).
- Room-relay resolution (NIP-65 ∪ naddr hints ∪ room relays tag) for all subscriptions.
- Resilient audio: cliff detector, JWT refresh, network-change recycle, auto-promote/demote.
- Speaking detection (energy-gated mic + remote level polling), participant action sheets,
  tabbed room view (Chat/Audience/Hands).
- `moq-nest.js` gains WebSocket fallback and per-participant level computation.
- Native MoQ track renamed for wire compatibility with web + reference impl.

### Unified NIP-78 preference blobs (Phase 0–4)
Consolidates scattered per-store preferences into two encrypted kind-30078 blobs:
- `nostr.blue/prefs` (NIP-44 to self via main signer) — settings, sidebar, reactions, AI creds.
- `nostr.blue/p2p` (NIP-44 to self via Mostro identity key) — Mostro settings, node config,
  bounded trade history with archival spillover.
- Dual-read migration (legacy + new), persistent cross-device sync subscriptions,
  quorum-EOSE fetch, and sunset markers on legacy loaders.

### Feeds infrastructure
New `src/feeds/` module (cursor, filter, ingestion, loader, ordering, outbox, pager, realtime,
repository, types) plus a rewritten `use_feed` hook — panic-safe task spawning and dispatcher
hardening throughout.

---

## Other additions
- **Topics system**: media replies, drafts, discover page, link previews, popular sidebar,
  NIP-73 web-URL hashtag migration.
- **Music explore**: new albums/artists/tracks/playlists/library routes + Wavlake explore service.
- **Muted words**: content filter applied across all feed views.
- **Weather dashboard**: redesigned cards + unit handling.
- **Publications**: NKBIP-01 e-tag section refs; all event kinds accepted.

## Removals
- **Blobbi virtual pet** subsystem removed entirely (~100 files, components/hooks/stores/routes).

## Bug fixes & hardening
- "Event not found" when opening feed and hot-post items; nevent links always embed author.
- Profile loading for follows, feed authors, and mentions (batched indexer fetching).
- Relay pool capped at 30 relays on WASM; outbox relays made evictable.
- WASM panics from `std::time::Instant` and `tokio::time::sleep` (replaced with platform-safe variants).
- Notification duplicates, content_filter panic, note-viewer thread fetching.
- Android Google sign-in, Drive upload, and multi-account support.
- NIP-78 relay-readiness gating, retry coverage, and logout cache cleanup.


src/stores/relay/scoped_subs.rs
```diff
@@ -0,0 +1,407 @@
...
+                    let _ = client.unsubscribe(&old_id).await;
+                });
+            }
+        }
+
+        // Subscribe with new config
+        let sub_id = if let Some(urls) = &config.relay_urls {
+            self.inner
+                .client
+                .subscribe_to(urls.clone(), config.filters.first().cloned().unwrap_or_default(), None)
```
goodmorningbitcoin, line 294: 🟠 **Bug: Only the first filter is used for subscriptions**

`config.filters.first().cloned().unwrap_or_default()` silently drops all filters after the first. The `SubConfig` struct accepts `Vec<Filter>` and `canonical_eq` compares the full vec, but the actual subscription only uses `filters[0]`. A multi-filter subscription like `[TextNote, Repost]` would only subscribe to `TextNote`, silently losing all reposts.

The SDK's `subscribe_to` and `subscribe` methods only accept a single `Filter`. To support multi-filter configs, either:
1. Loop and create one subscription per filter, or
2. Use `client.subscribe_many(vec![filter1, filter2], opts)` if the SDK supports it.

If multi-filter support isn't needed yet, document the constraint (e.g. add `debug_assert!(config.filters.len() <= 1, "..."))` or make `SubConfig.filters` a single `Filter` instead of a `Vec`.

src/stores/mostro/flow.rs
```diff
@@ -0,0 +1,967 @@
...
+                if inv.currency() != lightning_invoice::Currency::Bitcoin {
+                    return Err(format!(
+                        "Invoice is for {:?}, expected mainnet (Bitcoin)",
+                        inv.currency()
+                    ));
+                }
+                if let Some(expected) = expected_sats {
+                    if let Some(msats) = inv.amount_milli_satoshis() {
+                        let invoice_sats = msats.div_ceil(1000);
+                        if invoice_sats != expected && invoice_sats != 0 {
```
goodmorningbitcoin, line 83: 🟠 **0-sat invoice accepted as valid**

`invoice_sats != expected && invoice_sats != 0` — a zero-amount invoice passes the amount check even when `expected_sats` is non-zero. While this might be intentional for invoices where the amount is set later by the daemon, in the context of Lightning P2P trades, accepting a 0-sat invoice as matching an expected non-zero amount could allow a buyer to provide a malformed invoice that the daemon rejects later (after the trade index slot is already consumed).

If 0-sat invoices are legitimately used (e.g. amount-less invoices resolved by the daemon), add a comment explaining why. Otherwise, reject 0-sat invoices when `expected_sats` is provided: `if invoice_sats != expected { return Err(...) }`.

src/stores/mostro/keys.rs
```diff
@@ -0,0 +1,408 @@
...
+    /// Initialization or persistence failed.
+    Error(String),
+}
+
+/// Mostro key material plus the persistent state needed to derive per-trade keys.
+#[allow(dead_code)]
+#[derive(Clone, Debug)]
+pub struct MostroKeys {
+    /// 12-word BIP-39 mnemonic. Zeroized on drop to avoid lingering in
+    /// heap memory. Stored encrypted-at-rest per platform storage policy.
```
goodmorningbitcoin: ⚠️ **Misleading security comment: "Stored encrypted-at-rest"**

The comment says `Stored encrypted-at-rest per platform storage policy`, but `platform::storage` on WASM uses `localStorage`, which is plaintext and accessible to any JS running on the same origin (including XSS). On Android, `platform::storage` likely uses SharedPreferences (also plaintext).

The module doc at line 7 correctly says `plaintext, per design decision`, but this field comment contradicts it. Either:
1. Fix the comment to say "stored plaintext" (matching the module doc), or
2. Actually encrypt at rest using a key derived from the user's auth (e.g. NIP-44 to self)

For financial key material controlling Lightning trades, plaintext storage in localStorage is a real risk — an XSS attack would compromise the user's entire Mostro trade history and identity.

goodmorningbitcoin (commented):
## 🤖 Code Review (Hermes Agent)

Reviewed all new/changed code across Mostro P2P, Nests rebuild, Feeds infra, and utilities. Found issues across 15 files.

### Summary
- 🟠 **2 bugs** that could cause incorrect behavior (dropped filters, accepted 0-sat invoices)
- ⚠️ **6 warnings** (security, resource leaks, broken task lifecycle)
- 💡 **4 suggestions** (defense-in-depth, performance, robustness)

Inline comments below. Full summary in the follow-up comment.

src/feeds/outbox.rs
```diff
@@ -0,0 +1,461 @@
...
+                    .entry(relay_url.clone())
+                    .or_default()
+                    .push(filter);
+            }
+        }
+
+        // Step 5: Build indexer safety-net filters for unrouted authors
+        // AND as a general safety net (wisp sends full list to indexers)
+        let mut indexer_filters = Vec::new();
+        if !authors.is_empty() {
```
goodmorningbitcoin: ⚠️ **Indexer safety net sends ALL authors, not just unrouted**

Lines 255-267 send the **full** author list to `INDEXER_RELAYS`, not just the `unrouted` authors. For a Following feed with 500 follows, this creates `ceil(500/200) × 2 indexer relays = 6` additional subscriptions with 200-author filters each.

The wisp pattern this was ported from (`OutboxRouter.kt`) sends only unrouted authors to indexers as a fallback. Sending ALL authors doubles the relay load and filter complexity.

Consider: `if !unrouted.is_empty() { for chunk in unrouted.chunks(MAX_AUTHORS_PER_FILTER) { ... } }` — only send authors that couldn't be routed to their write relays.

src/feeds/outbox.rs
```diff
@@ -0,0 +1,461 @@
...
+}
+
+/// Builds per-relay filter targets for outbox-routed subscriptions.
+///
+/// One instance per feed scope. Uses a 5-minute TTL cache for author relay
+/// lists to avoid re-querying the database on every pagination.
+pub struct OutboxRouter {
+    client: Arc<Client>,
+    db: Arc<dyn FeedDatabase>,
+    relay_list_cache: Mutex<HashMap<PublicKey, CachedRelayList>>,
```
goodmorningbitcoin, line 101: ⚠️ **relay_list_cache is unbounded**

`relay_list_cache: Mutex<HashMap<PublicKey, CachedRelayList>>` grows without limit. Every author encountered in a feed gets a permanent entry. For a global firehose feed, this means thousands of entries accumulate with no eviction. The 5-minute TTL marks entries as stale (triggering a re-fetch) but never removes them.

Consider adding an LRU eviction policy or a max-entries cap (e.g. 1000 entries). When the cap is reached, evict the oldest entries first.

src/stores/mostro/client.rs
```diff
@@ -0,0 +1,2040 @@
...
+///
+/// This catches events that arrived while the app was closed or the
+/// subscription was unmounted. Without this, the user would miss trade
+/// updates that happened between sessions.
+///
+/// The 3-day window is required because gift-wrap envelopes have randomized
+/// `created_at` (see `nostr::nips::nip59::RANGE_RANDOM_TIMESTAMP_TWEAK =
+/// 0..172800` — up to 2 days back). Re-processed events are deduped via
+/// `dedup::SEEN_EVENTS`.
+pub async fn backfill_active_trades() {
```
goodmorningbitcoin, line 388: 💡 **backfill_active_trades has no concurrency guard**

`backfill_active_trades()` is an async fn that processes events sequentially, but nothing prevents two concurrent invocations (e.g. the live subscription's on-mount call and a `mostro-wake` push event arriving simultaneously). Two concurrent backfills would both fetch and process the same events, with the dedup LRU (`SEEN_EVENTS`) as the only guard — adding unnecessary load.

Consider an `AtomicBool` flag (`is_backfilling`) to short-circuit if a backfill is already in progress.

src/feeds/ingestion.rs
```diff
@@ -0,0 +1,162 @@
...
+//! ```
+
+use std::time::Duration;
+use instant::Instant;
+
+use nostr_sdk::Event;
+
+/// Default frame budget: 8ms on native, leaving ~8ms for rendering within
+/// a 16.6ms frame at 60fps.
+pub const DEFAULT_FRAME_BUDGET: Duration = Duration::from_millis(8);
```
goodmorningbitcoin, line 28: 💡 **DEFAULT_FRAME_BUDGET on WASM may be too aggressive**

The default budget is 8ms on native and 5ms on WASM (line 31). But WASM single-threaded execution also handles relay WebSocket I/O, signal reactivity, and rendering in that same thread. 5ms of event processing per frame may still cause jank on slower mobile devices.

Consider making the budget configurable per feed (e.g. 3ms for the global firehose, 8ms for a profile feed) or measuring actual frame times and adjusting dynamically.

src/stores/mostro/encrypted_attachment.rs
```diff
@@ -0,0 +1,362 @@
...
+        .map_err(|e| format!("legacy decryption failed: {e}"))
+}
+
+/// Phase 5.1 (C7): return the raw ECDH shared secret as the key.
+///
+/// Previously this applied a domain-separation hash
+/// (`SHA-256("mostro-chat-attachment-key-v1" || shared_secret)`), which
+/// was incompatible with mostro-cli/mostrix/mobile. The spec uses the
+/// raw ECDH shared secret directly.
+pub fn attachment_key_from_shared_secret(secret: &[u8; 32]) -> [u8; 32] {
```
goodmorningbitcoin, line 166: 💡 **Raw ECDH shared secret used directly as encryption key**

`attachment_key_from_shared_secret` returns the raw 32-byte ECDH shared secret as the ChaCha20Poly1305 key. This is documented as spec-compatible with mostro-cli/mostrix/mobile (line 160-168).

While this matches the reference implementations, raw ECDH without domain separation (e.g. HKDF or SHA-256 hash) is fragile — if the same keypair is ever reused for a different protocol, cross-protocol attacks become possible. The NIP-44 spec itself applies HKDF after ECDH.

Not a blocking issue since this matches the wire format, but worth documenting the security trade-off more prominently, or advocating upstream for a KDF.

src/components/qr_scanner.rs
```diff
@@ -0,0 +1,196 @@
...
+    // Find and decode QR codes using rqrr's PreparedImage API.
+    let mut prepared = rqrr::PreparedImage::prepare(gray);
+    let grids = prepared.detect_grids();
+
+    if grids.is_empty() {
+        return Err("No QR code found in the image".to_string());
+    }
+
+    // Decode the first grid found.
+    let (_meta, content) = grids[0]
```
goodmorningbitcoin: 💡 **Only first QR grid decoded; multi-QR images silently lose data**

Line 137 decodes only `grids[0]`. If a user captures an image containing multiple QR codes (e.g. a Lightning address + amount as separate codes), only the first is returned. Also, `grids[0].decode()` is not retried on failure — a partially-blurred first grid fails the whole scan even if a second grid is clear.

Consider: decode all grids and prefer the first that yields a valid `lnbc...` or Lightning address. Or at minimum, try `grids[i].decode()` for each until one succeeds.

goodmorningbitcoin (commented):
test batch 2

goodmorningbitcoin (commented):
test 0

goodmorningbitcoin (commented):
test 3

goodmorningbitcoin (commented):
Review batch 3: additional findings

public/moq-nest.js
```diff
@@ -186,6 +304,15 @@ window.nestAudioManager = window.nestAudioManager || {
...
+                        // detection. Matches Amethyst's `peakAmplitude` and
+                        // the local mic level computation in publishAudio.
+                        let peak = 0.0;
+                        for (let i = 0; i < buffer.length; i++) {
+                            const abs = Math.abs(buffer[i]);
+                            if (abs > peak) peak = abs;
+                        }
+                        subState.level = peak;
                         let audioBuffer;
                         while (bufferPool.length > 0) {
```
goodmorningbitcoin, line 317: 💡 **Buffer pool recycling doesn't check sampleRate**

The pool dimension check on line 319 compares `numberOfFrames` and `numberOfChannels` but not `sampleRate`. If a participant's encoder changes sample rate (e.g. from 48000 to 24000 during adaptive bitrate), a recycled `AudioBuffer` with the wrong sample rate would be used, causing audio artifacts (wrong playback speed).

Add `&& cached.sampleRate === sampleRate` to the dimension check.

goodmorningbitcoin:
## 🤖 Code Review Summary

Reviewed **414 files (+45,222 / −21,662)** across Mostro P2P, Nests rebuild, unified prefs, feeds infrastructure, and utilities. Here's the full finding inventory:

---

### 🟠 Bugs (2)

| File | Issue |
|------|-------|
| `scoped_subs.rs:294` | **Only first filter used for subscriptions** — `filters.first()` silently drops multi-filter configs |
| `flow.rs:83` | **0-sat invoice accepted** — `invoice_sats != 0` bypass allows zero-amount invoices to pass amount validation in financial code |

### ⚠️ Warnings (8)

| File | Issue |
|------|-------|
| `nip53.rs:1006` | **Admin kick has no explicit signature verification** — relies on SDK relay filtering; defense-in-depth gap for security-critical action |
| `keys.rs:58` | **Misleading "encrypted-at-rest" comment** — mnemonic is plaintext in localStorage on WASM, contradicted by module doc |
| `outbox.rs:255` | **Indexer safety net sends ALL authors** — not just unrouted, doubling relay load on indexer relays |
| `outbox.rs:97` | **relay_list_cache is unbounded** — global feed accumulates thousands of permanent entries with no eviction |
| `sw.js:2` | **Cache name hardcoded to 'v1'** — never versioned, so stale WASM/CSS assets accumulate after updates. The activate handler only deletes caches with *different* names, so old assets in the same key persist indefinitely |
| `nest_viewer.rs:366` | **Background tasks killed by init_for_room, use_hook won't re-spawn** — heartbeat, cliff detector, JWT refresh, and level poll permanently dead after room switch without page navigation |
| `moq-nest.js:103` | **Announce loop never cancelled** — async IIFE leaks after disconnect; `conn.announceTask = null` doesn't cancel the running coroutine |
| `moq-nest.js:317` | **Buffer pool doesn't check sampleRate** — wrong-sample-rate AudioBuffer reuse causes playback artifacts |

### 💡 Suggestions (4)

| File | Issue |
|------|-------|
| `client.rs:372` | **backfill_active_trades has no concurrency guard** — two concurrent invocations possible |
| `ingestion.rs:28` | **5ms WASM frame budget may be too aggressive** for slower mobile devices |
| `encrypted_attachment.rs:166` | **Raw ECDH without domain separation** — matches spec but fragile; NIP-44 itself uses HKDF |
| `qr_scanner.rs:137` | **Only first QR grid decoded** — multi-QR images silently lose data |

---

### ✅ What looks good
- Mostro trade state machine: monotonicity guard with snapshot/rollback on rejected transitions — solid
- NIP-44 encryption/decryption with legacy plaintext fallback — well-tested with round-trip and wrong-key tests
- NIP-06 key derivation — correct path, proper identity/trade-key separation, good test coverage
- Feeds cursor model: gap-proof backward pagination with per-relay state machine — clean port of amethyst's design
- Admin command authority check (freshness + dedup + host role) — good defense layers, just missing the signature verify
- Invoice validation: mainnet-only check, expiry check, amount check — thorough

---

*Reviewed by Hermes Agent • 15 inline comments posted across 3 review batches*

src/utils/nips/nip53.rs
```diff
@@ -827,6 +962,133 @@ pub fn nest_effective_status(
...
+        return None;
+    }
+    // Action: only honor `kick` (reference client doesn't implement mute/unmute).
+    let action = event
+        .tags
+        .iter()
+        .find(|t| t.as_slice().first().map(|s| s.as_str()) == Some("action"))
+        .and_then(|t| t.as_slice().get(1))
+        .map(|s| s.as_str());
+    match action {
```
goodmorningbitcoin, line 1006: 🟠 **Admin kick command: no explicit event signature verification**

`should_honor_admin_command` checks freshness, dedup, and signer authority (pubkey matches room host), but never calls `event.verify()`. While nostr-sdk verifies signatures on relay receipt in most configurations, defense-in-depth requires explicit verification for security-critical actions like kicking a user from a room.

An attacker who can inject a forged event (e.g. via a custom relay, a compromised relay, or a relay that doesn't verify) could craft a kind 4312 event with the host's pubkey and kick arbitrary users.

Add explicit signature verification before the authority check.

goodmorningbitcoin (commented):
Review follow-up: verified finding

public/moq-nest.js
```diff
@@ -75,6 +92,50 @@ window.nestAudioManager = window.nestAudioManager || {
...
 
+    /// Phase 4.1: Subscribe to the MoQ ANNOUNCE stream and cache the list of
+    /// announced participant pubkeys. Polled by Rust every 3s.
+    _startAnnounceLoop(publisherId) {
+        const conn = this._getConnection(publisherId);
+        if (!conn || !conn.connection) return;
+        if (conn.announceTask) return; // already running
+
+        const pubkeyRegex = /^[0-9a-f]{64}$/;
+        conn.announceTask = (async () => {
```
goodmorningbitcoin, line 103: ⚠️ **Announce loop never cancelled — leaks after disconnect**

`conn.announceTask` is an async IIFE running `for (;;) { await announced.next(); }`. The `disconnect()` function (line 463) sets `conn.announceTask = null` but doesn't actually cancel the running async function. The `announced()` async iterator will keep awaiting indefinitely until the connection object is garbage-collected.

Since `disconnect()` deletes the connection from the map (line 472), the old connection and its announce loop become orphaned. In practice, the GC may not collect it promptly if the `announced.next()` promise holds a reference chain back to the connection.

Consider: use `AbortController` or track the task and explicitly abort the iterator. Or at minimum, check `conn.state` inside the loop: `if (conn.state === 'disconnected') break;`.

goodmorningbitcoin (commented):
Review follow-up: verified finding

src/components/viewers/nest_viewer.rs
```diff
@@ -246,75 +343,433 @@ pub fn NestViewer(naddr: String) -> Element {
...
+    // Presence heartbeat (60s tick while joined).
     {
         let coord = room_coordinate.clone();
-        let is_joined_hb = is_joined;
-        let is_muted_hb = is_muted;
-        let is_publishing_hb = is_publishing;
-        let hand_raised_hb = hand_raised;
         use_hook(move || {
-            spawn(async move {
+            let task = spawn(async move {
```
goodmorningbitcoin, line 366: ⚠️ **Background tasks cancelled by init_for_room but use_hook won't re-spawn**

`use_hook` (lines 365-386) spawns the heartbeat task exactly once per component mount. When `nest_room_store::init_for_room()` is called (e.g. user navigates to a different room without leaving the page), it cancels the heartbeat, cliff detector, JWT refresh, and level poll tasks via `task.write().take().cancel()` (nest_room_store.rs:151-155).

But `use_hook` won't re-fire — it only runs once. After `init_for_room`, the heartbeat, cliff detector, JWT refresh, and level poll are all permanently dead. The user joins a new room with no heartbeat (presence stops updating after 10 min) and no speaking detection.

Fix: use `use_effect` keyed on the naddr or publisher_id instead of `use_hook`, so the tasks are re-spawned when the room changes. Or check inside the task loop whether the task's room matches the current room and re-register if not.

goodmorningbitcoin (commented):
Review follow-up: verified finding

src/components/viewers/nest_viewer.rs
```diff
@@ -246,75 +343,433 @@ pub fn NestViewer(naddr: String) -> Element {
...
+                // Local speaking-state hysteresis.
+                let mut speaking_until_secs: f64 = 0.0;
+                // Phase 3.7: per-remote-speaker hysteresis deadlines.
+                let mut remote_speaking_until: std::collections::HashMap<String, f64> =
+                    std::collections::HashMap::new();
+                const SPEAKING_THRESHOLD: f32 = 0.06;
+                const HYSTERESIS_SECS: f64 = 0.25;
+
+                loop {
+                    crate::platform::timer::sleep_ms(100).await;
```
goodmorningbitcoin, line 497: 💡 **Level poll runs at 100ms forever, even when not joined**

The level poll loop (lines 496-564) runs every 100ms unconditionally. When `is_joined` is false, it still wakes every 100ms, reads the signal, writes zero values, and continues. This wastes CPU cycles — 10 signal reads + writes per second doing nothing useful.

Consider using a longer sleep when not joined (e.g. 1s or 2s) and resuming 100ms polling only when `is_joined` becomes true. This is especially relevant on mobile where every wake-up drains battery.

goodmorningbitcoin (commented):
Review follow-up: verified finding

PR #344, opened 2026-07-01
0.8.19: NIPs spec registry, podcast NIP-F4, Mostro durability, radio management

patrickulrich:
Eight commits toward the **0.8.19** release. The headline change is a top-to-bottom
rework of the `/nips` page (replacing GitHub submodule docs with a hardcoded spec
registry and embedded markdown), plus new spec support (podcast NIP-F4, NIP-5A),
hardening of the Mostro P2P exchange, radio station management, and a pair of
Android-only bug fixes.

### `/nips` page overhaul
- **Replace GitHub submodules with a hardcoded registry** (`routes/nips/registry.rs`):
  `SUPPORTED_SPECS` is now the single source of truth for which NIPs/NUTs/BUDs/NKBIPs/
  Market specs nostr.blue implements. Drops `public/docs/{nips,nuts,blossom,market-spec,NKBIPs}`
  submodules, `.gitmodules`, submodule init/validation from CI, and the 455-line
  `services/github_nips.rs` (−1223 / +403 net).
- **Embed full spec markdown** via `include_str!` so detail pages render rich content
  offline instead of stub + upstream link. Adds per-spec notes under
  `routes/nips/content/`.
- **naddr-backed custom NIPs** (Pinboards, NIPs on Nostr, Music Tracks/Playlists) whose
  content is fetched live from relays; consolidated the seven old tabs into two
  (Our NIPs + Custom) with type-filter chips.
- **Resolve spec cross-links** (e.g. `[NIP-13](13.md)`) via a `spec_links` HTML post-pass —
  in-app routing for supported specs, upstream link for the rest. Repairs ~150 broken
  NUT references.
- **Correct NIP-29 attribution** (`/groups` is NIP-29, not NIP-28); updates README +
  AGENTS.md.

### Podcasts — NIP-F4 support
- Implement NIP-F4 (kinds 10154 / 54 / 10164 / 10054) alongside the existing legacy
  custom scheme, with **dual-read** so both schemas render. Subscriptions stay on
  NIP-51 kind 30003. F4 authorship verification (kind 10164 counter-claims) included.
- `PodcastAddr` enum (`Legacy` vs `F4`) threaded through player, library, cards, share,
  zap, vote; routing updated (`nevent` for kind 54, `naddr` for kind 10154).

### Mostro P2P exchange — durability & recovery
- **Role-aware upsert** (keyed on `(order_id, role)`) so self-takes no longer clobber the
  maker trade and orphan the order; self-take guard in `take_order`.
- **Mnemonic durability**: Mostro identity mnemonic is now backed up in the main NIP-78
  prefs blob (main-signer-encrypted), survives a localStorage wipe, and syncs across
  devices.
- **Global `use_mostro_session` hook** (app shell) so daemon GiftWrap replies are caught
  on every route, not just `/mostro`.
- `recover_order_by_id`, restore hardening, durable creation ledger, reactive My Trades.

### Radio
- Station **editing** (preserves the `d_tag` so edits replace, not duplicate),
  **overflow menu** (Edit / Broadcast / Copy ID / Delete), and **NIP-09 coordinate
  deletion** with confirm modal.
- Infinite-scroll pagination on radio home (until-timestamp paging, de-dup by coordinate).

### Content routing & viewers
- Route **kind 30030 (NIP-51 emoji set) to a new `EmojiPackViewer`** — it was previously
  misrouted to the Citation viewer via a bogus `30030..=30033` arm.
- Build out the stubbed **CitationViewer** for kinds 30–33 (NKBIP-03); add inline
  emoji-pack minicards for `nostr:naddr1...` of kind 30030.

### Android fixes
- **Weather geolocation**: widen the `navigator.geolocation` cfg gate from `web`-only to
  `any(web, mobile_platform)`; add the missing `ACCESS_FINE/COARSE_LOCATION` manifest
  perms that wry's permission check requires.
- **Copy/paste popups**: new `use_long_press` hook replaces blanket `contextmenu`
  `prevent_default()` in 4 components, restoring Android's native text-selection ActionMode.

### Other
- Auth: normalize stored pubkeys on session restore; clear corrupted storage entries.
- Calendar: NIP-09 deletion for events + live activities.
- Inline video: portrait detection via `imeta` dim tags with "Show More" collapse.
- One-time app-init guard, infinite-scroll sentinel backoff, Mostro double-borrow fix.


src/components/viewers/podcast_nostr_viewer.rs
```diff
@@ -198,6 +239,27 @@ fn PodcastDetailContent(props: PodcastDetailContentProps) -> Element {
...
                                 p { class: "text-muted-foreground mt-1", "{author}" }
                             }
+                            // NIP-F4 authors with verification status (kind 10164 counter-claims)
+                            if !f4_authors.is_empty() {
+                                div { class: "flex flex-col gap-1 mt-1",
+                                    for (author, verified) in &f4_authors {
+                                        div {
+                                            class: "flex items-center gap-1.5 text-xs text-muted-foreground",
+                                            span { class: "font-mono",
+                                                "{&author.pubkey[..8]}…",
```
goodmorningbitcoin: 🔴 **CRITICAL**: `&author.pubkey[..8]` panics if the string is shorter than 8 bytes. Data comes from untrusted relay `p` tags — a malicious relay sending `["p", ""]` causes a WASM panic that kills the app. Fix: `author.pubkey.get(..8).unwrap_or(&author.pubkey)` or use `.chars().take(8)`.

goodmorningbitcoin (commented):
test

goodmorningbitcoin (commented):
test

src/stores/mostro/restore.rs
```diff
@@ -860,6 +1303,119 @@ fn enrich_from_order_events() {
...
+    }
+
+    #[test]
+    fn test_derive_my_trade_index_finds_matching_key() {
+        let mk = test_keys(false);
+        let pk_at_3 = mk.get_trade_key_by_index(3).unwrap().public_key().to_hex();
+        let order = mostro_core::prelude::SmallOrder {
+            seller_trade_pubkey: Some(pk_at_3),
+            ..Default::default()
+        };
```
goodmorningbitcoin: 🔴 **CRITICAL**: Unchecked `i64 as u32` cast in `handle_last_trade_index`. `kind.trade_index()` returns `i64`. A negative value from the daemon wraps to `u32::MAX`, corrupting the trade-index counter permanently. The bounds-check guard exists in `handle_restore_data` (~line 490) but is missing here. Fix: `if remote_raw < 0 { log::warn!(...); return; }`

goodmorningbitcoin (commented):
test

src/stores/audio/music_library.rs
```diff
@@ -64,12 +69,18 @@ impl MusicLibraryItem {
...
+            TrackSource::Nostr { coordinate, .. } | TrackSource::Radio { coordinate, .. } => {
+                Some(Tag::custom(nostr_sdk::TagKind::a(), vec![coordinate.clone()]))
+            }
+            TrackSource::NostrPodcast { addr, .. } => match addr {
+                crate::stores::nostr_music::PodcastAddr::Legacy { coordinate, .. } => {
+                    Some(Tag::custom(nostr_sdk::TagKind::a(), vec![coordinate.clone()]))
+                }
+                // NIP-F4 episodes are regular (event-id addressed); persist via `e` tag.
+                crate::stores::nostr_music::PodcastAddr::F4 { event_id } => {
+                    Some(Tag::custom(nostr_sdk::TagKind::e(), vec![event_id.clone()]))
```
goodmorningbitcoin, line 81: 🟡 **WARNING**: NIP-F4 round-trip data loss. F4 episodes are persisted here as `e` tags, but `parse_library_event` does not parse `e` tags back. Saved F4 episodes are published to relays but silently dropped on reload. Fix: add an `e` tag case to `parse_library_event` that reconstructs the F4 episode.

src/stores/mostro/trade_store.rs
```diff
@@ -596,19 +596,36 @@ pub fn all_trades_for_daemon() -> Vec<Trade> {
...
+/// `take_order`). Previously upsert matched on `order_id` alone, so a taker
+/// record (from a self-take) silently overwrote the maker record —
+/// destroying the maker's handle and making the order vanish from My Trades.
+/// Requiring the role to match means a maker and taker record for the same
+/// order coexist as distinct entries instead of one clobbering the other.
 ///
 /// The change is reflected in the global signal and the local cache
 /// immediately. The NIP-78 publish happens separately via [`publish`].
 #[allow(dead_code)]
 pub fn upsert(trade: Trade) {
```
goodmorningbitcoin, line 614: 🟡 **WARNING**: `upsert` was updated to match by `(order_id, role)` — good. But `merge_remote_trades` (line ~800) still matches by `order_id` only. If both Maker and Taker records exist for the same order_id, `position()` finds only the first match, potentially leaving the other stale. Fix: add `&& t.role == remote.role` to the position lookup in `merge_remote_trades`.

src/stores/mostro/keys.rs
```diff
@@ -260,6 +309,36 @@ pub fn reset() {
...
+/// Restore a mnemonic from backup (cross-device / cleared-storage recovery).
+///
+/// Unlike [`import_mnemonic`], this does **not** reset the trade index — the
+/// index is recovered separately from the daemon's `LastTradeIndex` during
+/// the session restore (`sync_trade_index`). Only the mnemonic is
+/// persisted; keys are re-materialized reading the existing trade index.
+#[allow(dead_code)]
+pub fn restore_mnemonic(words: &str) -> Result<(), String> {
+    let trimmed = words.trim();
+    if trimmed.split_whitespace().count() != 12 && trimmed.split_whitespace().count() != 24 {
```
goodmorningbitcoin: ⚪ **NIT**: `split_whitespace().count()` called twice in the same condition. Extract to a local: `let word_count = trimmed.split_whitespace().count();`

goodmorningbitcoin (commented):
Review: final 3 comments (corrected line numbers against PR base)

goodmorningbitcoin (commented):
Review batch (inline comments 1)

src/stores/user_prefs/load.rs
```diff
@@ -318,9 +333,37 @@ pub fn apply_mostro_blob_to_signals(blob: &MostroPrefsBlob) {
         *crate::stores::mostro::trade_store::TRADES.write() = merged;
     }
 
+    // Creation ledger — merge remote entries into the local signal (union by
+    // (trade_index, role) / (order_id, role), keeping the newer). This
+    // restores the durable "orders I own" handle across devices.
+    if !blob.creation_ledger.is_empty() {
```
goodmorningbitcoin, line 347: 🟡 **WARNING**: `apply_mostro_blob_to_signals` directly writes the merged creation ledger to the `CREATION_LEDGER` signal (line 360) but does NOT call `creation_ledger::bump_version()`. The `CREATION_LEDGER_VERSION` signal exists so reactive consumers (the app-shell persistence watcher) can subscribe without cloning the Vec. Without the version bump, the watcher won't re-publish the Mostro blob, so remote-merged ledger entries may be lost on the next local-only snapshot publish. Fix: after the `*CREATION_LEDGER.write() = local;` line, call `crate::stores::mostro::creation_ledger::bump_version()` — but note `bump_version` is currently private; either make it `pub(crate)` or call `creation_ledger::append()` per entry instead of direct signal manipulation.

src/stores/user_prefs/load.rs
```diff
@@ -198,6 +198,21 @@ pub fn apply_blob_to_signals(blob: &UserPrefsBlob, source: BlobSource) {
         *crate::stores::mostro::nip78::P2P_TERMS_VERSION_ACCEPTED.write() = Some(version);
     }
 
+    // Mostro mnemonic — restore from the (main-signer-encrypted) backup when
+    // localStorage lacks it or differs. This is the cross-device /
+    // cleared-storage recovery path: the main blob rehydrates the Mostro
+    // identity, after which the Mostro blob + daemon restore can proceed
+    // with the correct identity. `restore_mnemonic` preserves the existing
+    // trade index (the daemon's LastTradeIndex syncs it during restore).
+    if let Some(ref remote_mnemonic) = blob.mostro_mnemonic {
```
goodmorningbitcoin, line 207: 🟡 **WARNING**: `apply_blob_to_signals` silently overwrites the local Mostro mnemonic when the remote (relay) value differs, with no user confirmation. If a user imported a different mnemonic on this device (via `import_mnemonic`), this code would revert it to the old relay value on the next blob load. While the blob is encrypted to self (trustworthy source), 'last relay write wins' can cause confusion if the user intentionally changed their Mostro identity locally. Consider logging when this path is taken (currently only logs on error) or surfacing a notification to the user so they understand their local identity was replaced.

src/stores/user_prefs/sidecar.rs
```diff
@@ -56,6 +56,9 @@ pub fn snapshot_main_blob() -> UserPrefsBlob {
     let reactions = crate::stores::reactions_store::PREFERRED_REACTIONS.read().clone();
     let notifications_checked_at =
         *crate::stores::ui::notifications::NOTIFICATIONS_CHECKED_AT.read() as u64;
+    // Back up the Mostro mnemonic (if keys are ready) so the Mostro identity
+    // survives a localStorage wipe and syncs across devices.
+    let mostro_mnemonic = crate::stores::mostro::try_get().map(|k| k.mnemonic.0.clone());
```
goodmorningbitcoin, line 61: 🔵 **SUGGESTION**: `snapshot_main_blob` extracts the mnemonic from `ZeroizeString` into a plain `String` (line 61: `k.mnemonic.0.clone()`). The `ZeroizeString` wrapper exists to zeroize sensitive material on drop, but the extracted `String` in the `UserPrefsBlob` will NOT be zeroized — it sits in heap memory until the blob struct is dropped (after async publish). For a BIP-39 mnemonic, consider using a `SecretString` or at minimum documenting that the mnemonic lives in the blob in plaintext. The NIP-44 encryption protects it on the wire, but the in-memory exposure window is widened.

src/stores/mostro/restore.rs
```diff
@@ -753,6 +906,296 @@ fn derive_counterparty(
...
+fn derive_my_trade_index(
+    mostro_keys: &keys::MostroKeys,
+    order: &mostro_core::prelude::SmallOrder,
+) -> Option<u32> {
+    if mostro_keys.privacy_mode {
+        return None;
+    }
+    let buyer = order.buyer_trade_pubkey.as_deref();
+    let seller = order.seller_trade_pubkey.as_deref();
+    for idx in 1..mostro_keys.trade_index {
```
goodmorningbitcoin, line 952: 🔵 **SUGGESTION**: `derive_my_trade_index` scans ALL trade indices from 1 to `trade_index`, deriving a full BIP-32 key for each and comparing pubkeys. For a user with a high trade index (e.g. 1000+), this is 1000 BIP-32 derivations per missing order during `merge_small_orders` create-on-miss. This runs synchronously on the main thread. Consider limiting the scan range or caching derived trade pubkeys. In practice trade_index is typically small (<100), so this is low priority.

src/stores/mostro/restore.rs
```diff
@@ -253,29 +281,123 @@ async fn request_restore_inner() -> Result<(), String> {
...
+/// it can silently drop the daemon's reply even though the event is stored
+/// on the relay. This fetches it explicitly.
+///
+/// Gift-wrap `created_at` is randomized (NIP-59), so we deliberately do NOT
+/// use `.since(...)` — a bounded `limit` retrieves the most recent daemon
+/// replies addressed to the identity key, and we process any
+/// restore/orders/last-trade-index payload among them. Processing is
+/// idempotent: re-handling `RestoreData` re-upserts the same trades and
+/// flips the stage to `Done`.
+async fn fetch_restore_reply_fallback(identity_pk: PublicKey) -> bool {
```
goodmorningbitcoin, line 303: 🔵 **SUGGESTION**: `fetch_restore_reply_fallback` fetches up to 50 GiftWrap events addressed to the identity key and processes ALL of them through `handle_restore_event`. While `handle_restore_message` only acts on RestoreSession/LastTradeIndex/Orders actions (ignoring others), processing 50 gift wraps involves NIP-44 decryption for each, which is CPU-intensive on WASM. Consider adding early-exit logic: once a RestoreData or LastTradeIndex has been successfully processed and the stage is `Done`, skip remaining events. Currently it processes all 50 even after finding the reply.

src/routes/nips/home.rs
```diff
@@ -1,170 +1,148 @@
...
-    let mut nuts = use_signal(Vec::<DocSpec>::new);
-    let mut buds = use_signal(Vec::<DocSpec>::new);
-    let mut nkbips = use_signal(Vec::<DocSpec>::new);
-    let mut market_spec_content = use_signal(|| None::<String>);
     let mut loading = use_signal(|| false);
+    let mut loading_more = use_signal(|| false);
     let mut has_more = use_signal(|| true);
     let mut oldest_timestamp = use_signal(|| None::<u64>);
     let mut error = use_signal(|| None::<String>);
+    let client_initialized = *nostr_client::CLIENT_INITIALIZED.read();
```
goodmorningbitcoin, line 69: 🟡 **WARNING**: `client_initialized` is read as a plain `bool` outside any reactive scope (line 66) and then captured by `move` in the `use_effect` at line 69. The effect only re-runs when `active_tab` changes. If `CLIENT_INITIALIZED` transitions from false→true while the user is already on the Custom tab, the effect won't re-run — Custom NIPs will not load until the user switches tabs and back. Fix: read `CLIENT_INITIALIZED` inside the effect closure so Dioxus subscribes the effect to the signal.

src/routes/nips/home.rs
```diff
@@ -1,170 +1,148 @@
...
-            loading.set(false);
+            loading_more.set(false);
         });
     };
-    let sentinel_id = use_infinite_scroll(load_more, has_more, loading);
-    // Custom NIP relay search
+    let sentinel_id = use_infinite_scroll(load_more, has_more, loading_more);
+
+    // Custom-tab relay search.
     use_effect(move || {
```
goodmorningbitcoin, line 155: 🟡 **WARNING**: Custom NIP relay search spawns an async request on every `search_input` change with no debouncing. Concurrent searches can complete out of order — a slow query for 'abc' could overwrite results from a newer 'abcdef' query. Fix: add a request-ID guard or debounce (300ms timer before firing relay search).

src/components/rich_content/mentions.rs
```diff
@@ -642,8 +645,8 @@ pub fn NaddrMentionRenderer(mention: String) -> Element {
                         }
                     }
                 }
-                PODCAST_EPISODE => {
-                    if let Ok(episode) = parse_podcast_episode(&event) {
+                PODCAST_EPISODE | PODCAST_EPISODE_F4 => {
```
goodmorningbitcoin, line 648: 🟡 **WARNING**: F4 episode minicards link to `Route::PodcastNostrDetail` (the show page) instead of `Route::PodcastNostrEpisodeDetail`. For NIP-F4 episodes (kind 54), `naddr_for_link` is an episode identifier, not a show coordinate. `PodcastNostrDetail`'s `fetch_nostr_podcast_metadata` will fail to parse it. Fix: route to `PodcastNostrEpisodeDetail`.

src/components/reaction/button.rs
```diff
@@ -89,6 +89,95 @@ pub fn ReactionButton(props: ReactionButtonProps) -> Element {
...
+            }
+        }),
+        DEFAULT_LONG_PRESS_MS,
+    );
+
+    // Wrap touchstart to capture coordinates before forwarding to the hook.
+    let on_touch_start_wrapped = move |e: TouchEvent| {
+        if let Some(touch) = e.touches().first() {
+            let c = touch.client_coordinates();
+            touch_anchor.set(Some((c.x, c.y)));
```
goodmorningbitcoin, line 181: ⚪ **NIT**: `touch_anchor` signal is written on every `touchstart` on web builds but never read (long-press is no-op on web). Harmless but wastes a signal write per touch event. Consider gating behind `#[cfg(not(feature = "web"))]`.

src/routes/nips/nip_detail.rs
```diff
@@ -162,6 +94,7 @@ pub fn NipDetail(nip_id: String) -> Element {
                             .collect();
                         nip_title.set(title);
                         nip_content.set(Some(event.content.clone()));
+                        spec_source.set(Some(registry::SpecType::Nip));
```
goodmorningbitcoin: 🔵 **SUGGESTION**: `spec_source` hardcoded to `Nip` for all custom NIPs (kind 30817). If custom NIP markdown contains a bare `04.md` link referencing a NUT-04, it would be incorrectly rewritten to NIP-04's upstream URL. Fix: set `spec_source` to `None` for custom NIPs to disable link rewriting.

goodmorningbitcoin (commented):
Review batch (inline comments 2)

goodmorningbitcoin (commented):
Review batch (inline comments 3)

src/routes/nips/home.rs
```diff
@@ -1,170 +1,148 @@
...
-    let mut market_spec_content = use_signal(|| None::<String>);
     let mut loading = use_signal(|| false);
+    let mut loading_more = use_signal(|| false);
     let mut has_more = use_signal(|| true);
     let mut oldest_timestamp = use_signal(|| None::<u64>);
     let mut error = use_signal(|| None::<String>);
+    let client_initialized = *nostr_client::CLIENT_INITIALIZED.read();
+
+    // Custom-tab data loading (relay fetch).
     use_effect(move || {
```
goodmorningbitcoin, line 72: 🔵 **SUGGESTION**: `use_effect` resets error/loading signals on every tab switch, and re-fetches custom NIPs from relays each time. Consider caching custom NIPs data to avoid re-fetching on every tab switch.

src/components/viewers/radio_viewer.rs
```diff
@@ -50,15 +62,174 @@ pub fn RadioViewer(naddr: String) -> Element {
         }
     });
     rsx! {
```
goodmorningbitcoin, line 64: 🔵 **SUGGESTION**: Root `div` onclick handler closes menu on any click within the viewer. Clicks on child elements (play button, stream selector) bubble up and also trigger the close. No `stop_propagation()` on child handlers. Behavior appears intentional but undocumented — worth a comment.

src/components/rich_content/mod.rs
```diff
@@ -449,6 +485,13 @@ fn InlineVideoPlayer(url: String) -> Element {
     };
     let video_id = format!("iv-{}", url_hash);
     let mut captured_poster: Signal<Option<String>> = use_signal(|| None);
+    let is_tall_from_imeta = match (imeta_width, imeta_height) {
```
goodmorningbitcoin, line 488: 🔵 **SUGGESTION**: `InlineVideoPlayer`'s `expanded` signal persists when the video URL changes (component reused in scrolling feed). Fix: add `use_effect` that resets `expanded` when `url` prop changes.

src/routes/address_viewer.rs
```diff
@@ -405,14 +415,15 @@ fn dispatch_naddr(kind: u16, naddr: String, coord: &Nip19Coordinate) -> std::res
             let identifier = coord.coordinate.identifier.to_string();
             Ok(AddressState::Wiki { npub, identifier })
         }
-        30030..=30033 => Ok(AddressState::Citation { naddr }),
+        30030 => Ok(AddressState::EmojiPack { naddr }),
```
goodmorningbitcoin, line 418: ⚪ **NIT**: Kind 30030 correctly routed to EmojiPack, but kinds 30031-30033 are now unhandled and fall through to 'not yet supported' error. Consider adding a catch-all fallback to `AddressState::Article` for unknown addressable kinds.

src/components/viewers/calendar_event_viewer.rs
```diff
@@ -540,6 +546,17 @@ pub fn CalendarEventViewer(naddr: String, from: Option<String>) -> Element {
                                                     },
                                                     if *is_copying.read() { "Copying..." } else { "Copy Event ID" }
                                                 }
+                                                if is_own {
+                                                    button {
+                                                        class: "w-full text-left px-4 py-2 hover:bg-accent transition-colors flex items-center gap-2 text-sm text-red-500",
+                                                        disabled: *is_deleting.read(),
```
goodmorningbitcoin, line 552: 🔵 **SUGGESTION**: Delete spawn closure resets `is_deleting` signal, but if user navigates away before spawn completes, the signal write targets a stale component instance. Dioxus drops it gracefully, but UI may briefly show 'Deleting...' state.

src/components/mostro/take_mostro_button.rs
```diff
@@ -15,6 +15,11 @@ pub fn TakeMostroButton(order: P2POrder) -> Element {
     let range_amount = use_signal(String::new);
     let is_range = matches!(order.fiat_amount, FiatAmount::Range { .. });
     let is_sell = order.order_type == OrderType::Sell;
+    // Disable taking if the current user already owns this order as maker.
+    // `take_order` blocks self-takes server-side too, but disabling here is
+    // better UX (no error toast) and avoids corrupting local state.
+    let owns_order = crate::stores::mostro::find_by_order_id(&order.order_id)
```
goodmorningbitcoin, line 21: 🔵 **SUGGESTION**: `owns_order` computed during render via `find_by_order_id`. Verify that this function reads reactively from a GlobalSignal (like `TRADES`) so the button re-renders correctly when trades change. If it reads from a non-reactive cache, `owns_order` could be stale.

src/utils/audio/podcast.rs
```diff
@@ -412,8 +474,305 @@ pub fn parse_podcast_metadata(event: &Event) -> Result<PodcastMetadata, String>
...
+        return Err(format!(
+            "Expected kind {}, got {}",
+            KIND_F4_EPISODE,
+            event.kind.as_u16()
+        ));
+    }
+    let pubkey = event.pubkey.to_hex();
+    let title =
+        get_tag_value(event, "title").ok_or("Missing required 'title' tag for NIP-F4 episode")?;
+    let audio_url = get_tag_value(event, "audio")
```
goodmorningbitcoin, line 553: 🟡 **WARNING**: SECURITY — No URL scheme validation on audio_url: The `audio_url` field (and similarly `image`, `chapters_url`, `transcripts[].url`, `soundbites[].url`, `funding[].url`) are parsed from untrusted relay event tags and stored as raw strings without validating the URL scheme. A malicious relay could inject `javascript:`, `data:`, or `file:` URIs. When these are used in `<audio src>` or `` elements, a `javascript:` URL won't execute in media tags, but `data:` URLs can be used for data exfiltration or XSS in certain rendering contexts. The `audio_url` is used for playback, and `image`/`thumbnail` fields are rendered in `` tags. Fix: Add a validation function (the codebase already has `crate::utils::validation::is_valid_http_url`) and filter or reject non-http(s) URLs during parsing.

src/platform/geolocation.rs
```diff
@@ -32,13 +37,13 @@ pub async fn get_current_position() -> Result<(f64, f64), String> {
...
-    #[cfg(not(feature = "web"))]
+    #[cfg(not(any(feature = "web", feature = "mobile_platform")))]
     {
         Err("Geolocation only available on web".to_string())
     }
 }
 
-#[cfg(feature = "web")]
+#[cfg(any(feature = "web", feature = "mobile_platform"))]
 pub fn start_watch_position(_callback_id: &str) {
```
goodmorningbitcoin, line 63: 🟡 **WARNING**: SINGLE-GLOBAL WATCH ID CONFLICT: `start_watch_position` stores the watch ID in `window._geoWatchId`, a single global JS variable. If two components/features call `start_watch_position` with different callback IDs, the first caller's watch is silently cleared by the second caller's `navigator.geolocation.clearWatch(window._geoWatchId)` call. The first feature stops receiving position updates with no error. Fix: Use a map of callback_id → watch_id (e.g., `window._geoWatchIds = {}`) or document that only one concurrent watch is supported.

src/components/viewers/podcast_nostr_viewer.rs
```diff
@@ -438,4 +522,44 @@ async fn fetch_nostr_episodes(
...
+/// For each author claimed in a podcast's kind 10154 `p` tags, fetches the
+/// author's own kind 10164 event and checks it lists the podcast's pubkey.
+/// Returns `(author, verified)` pairs. Failures (offline, no event) are treated
+/// as unverified rather than errors so the UI still renders.
+async fn verify_authors(
+    podcast_pubkey: &str,
+    authors: &[podcast::PodcastAuthor],
+) -> Vec<(String, bool)> {
+    let mut results = Vec::with_capacity(authors.len());
+    for author in authors {
```
goodmorningbitcoin: 🟡 **WARNING**: PERFORMANCE — Sequential authorship verification: `verify_authors` iterates through each author and sequentially `await`s `fetch_events_aggregated` per author with a 5-second timeout each. For a podcast with N authors, worst-case latency is N × 5s (e.g., 5 authors = 25s). This runs inside `spawn` on the UI thread. Fix: Use `futures::future::join_all` to fetch all author kind-10164 events concurrently, or use a single filter with multiple `.authors(...)` to batch the fetch into one relay round-trip.

src/components/media/lightbox.rs
```diff
@@ -7,6 +7,9 @@ use dioxus_core::use_drop;
 #[cfg(feature = "web")]
 use gloo_events::EventListener;
 
+static VIEWPORT_SIZE: GlobalSignal<(f64, f64)> = Signal::global(|| (1200.0, 800.0));
```
goodmorningbitcoin, line 14: 🔵 **SUGGESTION**: DESIGN — `VIEWPORT_SIZE` and `VIEWPORT_KNOWN` changed from component-local `use_signal` to `Signal::global` (GlobalSignal). This means viewport state persists for the entire app lifecycle and is shared across all instances. While MediaLightbox is effectively a singleton, GlobalSignals have different reactivity semantics than local signals — every component that reads them re-renders on any write. There is no comment explaining why this change was necessary. If the motivation is surviving unmount/remount cycles, consider documenting this. If not intentional, prefer the original local signals to avoid unnecessary global reactivity.

goodmorningbitcoin (commented):
Review batch (inline comments 4)

src/utils/audio/radio.rs
```diff
@@ -689,3 +705,28 @@ mod tests {
         assert_eq!(parsed_d_tag, d_tag);
     }
 }
+
+/// Delete a radio station by publishing a NIP-09 deletion request.
+/// Uses coordinate-based deletion (a-tag) so relays remove all versions.
+pub async fn delete_radio_station(coordinate: &str) -> Result<String, String> {
```
goodmorningbitcoin, line 719: 🔵 **SUGGESTION**: MISSING AUTHORIZATION CHECK: `delete_radio_station(coordinate: &str)` accepts an arbitrary coordinate string, parses it, signs a NIP-09 deletion request, and publishes it without verifying that the coordinate belongs to the current user. While relays should reject deletion requests for events not owned by the signer, some relays may have bugs. The radio viewer UI does check `is_own` before showing the delete button (radio_viewer.rs line 2531), but the utility function itself has no guard. Fix: Add an optional ownership check inside `delete_radio_station` that compares `coord.public_key` to `auth_store::get_pubkey()` and returns an error on mismatch.

goodmorningbitcoin (commented):
Review batch (inline comments 5)

src/stores/mostro/creation_ledger.rs
```diff
@@ -0,0 +1,122 @@
...
+/// Append or update a ledger entry. Dedup is by `(trade_index, role)` when
+/// the index is known (so a placeholder `maker-{N}` is migrated in-place to
+/// its real UUID on ACK), otherwise by `(order_id, role)`. Newest first;
+/// bounded to [`MAX_LEDGER_ENTRIES`].
+#[allow(dead_code)]
+pub fn append(entry: CreationLedgerEntry) {
+    let mut list = CREATION_LEDGER.write();
+    let role = entry.role;
+    let idx = entry.trade_index;
+    if let Some(existing) = list.iter_mut().find(|e| {
```
goodmorningbitcoin, line 81: 🔵 **SUGGESTION**: In `append`, the dedup condition `(idx.is_some() && e.trade_index == idx && e.role == role) || (e.order_id == entry.order_id && e.role == role)` has a subtle issue: when `idx.is_some()` but the index does NOT match the existing entry, it falls through to the `order_id` match. If a placeholder entry has order_id `maker-5` with `trade_index=Some(3)`, and a new entry arrives with `order_id`=`maker-5` but `trade_index=Some(4)` (different index), the OR's second arm matches on order_id and overwrites with the wrong trade_index. This edge case could arise if the same placeholder is reused with different indices. Consider requiring BOTH conditions match when the index is present: if `idx.is_some()`, match only on `(trade_index, role)`; if `idx.is_none()`, match only on `(order_id, role)`.

src/hooks/use_long_press.rs
```diff
@@ -0,0 +1,123 @@
...
+    impl FnMut(TouchEvent),
+) {
+    // Generation counter: bumped on every cancellation, so a stale timer can
+    // detect it was superseded by comparing its captured token to the current
+    // value.
+    let mut generation = use_signal(|| 0u32);
+
+    // Capture the callback into a Signal so the closures can read it without
+    // capturing non-`Copy` state directly.
+    let callback = use_signal(move || on_long_press);
```
goodmorningbitcoin: 🔵 **SUGGESTION**: `use_signal(move || on_long_press)` captures the callback once at initial render. If the callback changes on subsequent renders (e.g., captures different state), the stale callback fires. Mitigated by Dioxus key-based reconciliation today, but fragile. Fix: update the callback on each render.

src/hooks/use_mostro_session.rs
```diff
@@ -0,0 +1,188 @@
...
+/// always-mounted root `Layout` component (`routes/mod.rs`). Safe to call
+/// before login / before Mostro keys exist — it no-ops (filter `None`)
+/// until `try_get_mostro_keys()` and `try_get_node_config()` both return
+/// `Some`.
+#[allow(dead_code)]
+pub fn use_mostro_session() {
+    // Dedup buffer for events we've already processed. A persistent signal
+    // (not a local) so it survives re-subscriptions triggered by filter
+    // rebuilds without re-processing the backfill.
+    let mut seen_events: Signal<HashSet<EventId>> = use_signal(HashSet::new);
```
goodmorningbitcoin, line 49: 🟡 **WARNING**: MEMORY LEAK: `seen_events: Signal<HashSet<EventId>>` grows unboundedly for the lifetime of the app session. Every incoming Mostro daemon event (GiftWrap/NIP-44) inserts its `EventId` into this set and entries are never evicted. For an always-mounted global subscription that runs continuously, this HashSet will accumulate thousands of 32-byte EventIds over hours/days, steadily consuming WASM heap memory. Fix: Use a bounded LRU cache (the codebase already depends on `lru::LruCache`) or periodically clear entries older than N minutes. A capacity of ~500-1000 events would be sufficient for dedup purposes.

src/components/viewers/emoji_pack_viewer.rs
```diff
@@ -0,0 +1,257 @@
...
+                                "{button_label}"
+                            }
+                        })
+                    }
+
+                    // Emoji grid
+                    div { class: "grid grid-cols-4 sm:grid-cols-6 md:grid-cols-8 gap-3 pt-2",
+                        for emoji in p.emojis.iter() {
+                            div {
+                                key: "{emoji.shortcode}",
```
goodmorningbitcoin: ⚪ **NIT**: Emoji grid uses `key: "{emoji.shortcode}"`. Duplicate shortcodes in user-generated NIP-30 packs cause Dioxus key-collision warnings. Fix: use `key: "{idx}-{emoji.shortcode}"`.

src/routes/nips/spec_links.rs
```diff
@@ -0,0 +1,197 @@
...
+            };
+            let href = &hcaps["href"];
+            let Some((new_href, external)) = resolve_spec_href(href, source) else {
+                // Not a relative spec-file reference — leave it untouched.
+                return caps[0].to_string();
+            };
+            // Swap the href value, preserving all other attributes (e.g. the
+            // `rel="noopener noreferrer"` ammonia already added).
+            let replaced = format!("href=\"{new_href}\"");
+            let mut new_attrs = attrs.replace(&hcaps[0], &replaced);
```
goodmorningbitcoin: ⚪ **NIT**: `String::replace` replaces ALL occurrences of the matched href in the attribute string. Safe because ammonia sanitizes before this runs, but semantics are broader than the intended single replacement.

goodmorningbitcoin (commented):
Review batch (new files)

goodmorningbitcoin:
## Hermes Agent Review — PR #344 (0.8.19)

**39 findings: 3 critical, 13 warnings, 19 suggestions, 7 nits** across 3 parallel review groups.

30 inline comments posted below. 9 additional findings on code outside the diff hunks:

### 🔴 Critical (pre-existing, not introduced by this PR)
- **`restore.rs:617`** — Unchecked `i64 as u32` cast in max-trade-index computation. One negative index from daemon → `max()` picks the wrapped value → `sync_trade_index` sets `trade_index = u32::MAX`, permanently breaking trade-key allocation. The bounds-check guard exists at line ~498 but is missing here. Fix: `.filter(|o| o.trade_index >= 0).map(|o| o.trade_index as u32).max()`
- **`restore.rs:642`** — Same class of bug in `handle_last_trade_index`. `kind.trade_index() as u32` with no negative check. Fix: `if remote_raw < 0 { return; }`

### 🟡 Warnings (pre-existing / unchanged code)
- **`geolocation.rs:24`** — No timeout on `getCurrentPosition`. If user ignores the permission prompt, the future hangs forever. Wrap in `Promise.race` with a timeout.
- **`podcast.rs:938`** — V4V split inconsistency: tag parser drops recipient on missing split (`slice.get(4)?.parse().ok()?`), JSON parser defaults to 100 (`.unwrap_or(100)`). Fix: use consistent fallback in both parsers.

### 🔵 Suggestions (unchanged code)
- **`keys.rs:98`** — `wrapping_add(1)` on trade index could wrap to 0 (identity key derivation path) at `u32::MAX`. Use `saturating_add(1)`.
- **`radio_station_new.rs:497`** — Stream URLs from user input pushed to Nostr tags without validation. `is_valid_http_url` helper exists elsewhere.
- **`radio.rs:41`** — `StreamFormat::from_url` uses `.contains(".mp3")` — matches directory names, not just extensions.
- **`radio.rs:350`** — `fetch_station_by_naddr` retries 3× with 4s sleeps (12s worst case) with no progress feedback.

### ⚪ Nit
- **`use_infinite_scroll.rs:71`** — `log::info!` used extensively for debug-level diagnostics. Downgrade to `log::debug!`.

---
*Reviewed by Hermes Agent — 3 parallel subagent groups. [Full JSON findings](https://github.com/patrickulrich/nostr.blue/pull/344#issuecomment)*

PR #346, opened 2026-07-10
0.8.20: Routstr AI provider, Mostro trade durability, profile loading perf, cross-platform voice

patrickulrich:
Five commits toward the **0.8.20** release. Headline changes are a new built-in
AI provider (Routstr), Mostro trade durability/recovery hardening, a profile
page loading optimization with multi-source streaming, cross-platform voice
message playback, and a media upload refactor separating image compression from
generic file uploads.

### Routstr — built-in AI provider
- Adds [Routstr](https://routstr.com), a decentralized Bitcoin-powered
  OpenAI-compatible AI marketplace, as a second built-in provider alongside PPQ.
- New `ProviderAuth::Routstr` variant + `routstr_api_key` field on
  `AiProviderState` (serde-default for backward compat); `routstr_provider()`
  constructor as the second built-in in `resolve_providers`.
- **`src/services/routstr.rs`** (new): balance, Lightning invoice + status
  polling, Cashu topup, key creation from cashu token, refund. Defensive JSON
  parsing with redacted error bodies (mirrors `ppq.rs`). Endpoint shapes verified
  against routstr-core server source (`/lightning/invoice/{id}/status`, `{msats}`
  topup, string sats refund, `{detail}` error envelope).
- **`src/components/routstr_settings_panel.rs`** (new): API key entry with
  auto-conversion of `cashuA...` tokens to `sk-...` keys, balance display
  (msats→sats) with refresh, Lightning topup with QR + auto-poll (5s / 60
  attempts), Cashu token topup, refund/withdraw to copyable Cashu token.
- AI chat route: renamed `ppq_blocked` → `needs_setup` (10 occurrences,
  generalized copy); type-dispatched setup gates per provider — `PpqSetupGate`
  (account creation), `RoutstrSetupGate` (inline Lightning key creation with QR
  + auto-poll), `ProviderSetupGate` (generic for custom providers).

### Mostro — durable trade recovery & orphan resilience
- **create_order**: upserts a placeholder trade *before* sending the daemon
  message so `use_mostro_session` covers it during the ACK wait (survives a
  `Lagged` broadcast); `RecvError::Lagged` treated as recoverable and continues;
  placeholder removed on send failure or daemon rejection.
- **trade_detail**: "Recover from daemon" button when no local record exists
  (calls `recover_order_by_id`); real-UUID Pending trades past 30s transition to
  slow reconcile instead of being orphaned out from under the user.
- **my_trades**: self-heals an empty trade list from the NIP-78 blob, the
  creation ledger, and `RestoreSession` on mount.
- **cleanup**: creation-ledger trades exempt from the orphan sweep; orphan
  thresholds widened (120→600s, bond grace 180→900s); `is_live_on_board`
  broadened to any non-terminal status so active InProgress orders aren't swept.
- **take_mostro_button / order_card**: self-take disabled using both TRADES +
  creation ledger (survives a TRADES cache wipe); "Yours" badge and "Open Trade"
  shortcut for orders owned via either source.
- **client**: `AddInvoice` now handles the `Order` payload and preserves
  `PaymentFailed`; role-aware bond step (Maker→`WaitingMakerBond`,
  Taker→`WaitingTakerBond`); `ensure_node_relays_connected` in
  `resolve_effective_pow` so cold-start PoW fetch doesn't silently fail.
- **node_config**: `lnd_node_uri` → `lnd_uris` to match the daemon's tag name.
- **Mention autocomplete fix**: `update_dropdown_position` was double-counting
  scroll offset (adding `scroll_y()`/`scroll_x()` to coords already
  viewport-relative from `get_bounding_client_rect()`), shoving the `@`-mention
  dropdown below the viewport in scrolled reply/comment modals.

### Profile page — multi-source streaming & indexer routing
- **Metadata (kind 0)**: race dedicated indexer relays vs outbox path; first
  result wins (indexers <500ms vs 5–10s outbox). Relay resolution runs
  non-blocking in background instead of gating the metadata fetch.
- **Posts (tab events)**: replace blocking `join!` with multi-source streaming
  merge (connected pool + SDK gossip + optional outbox) via an mpsc channel with
  EventId dedup; events paint progressively through `DebouncedCollector` instead
  of waiting for full EOSE. New **`src/stores/nostr_client/streaming.rs`**
  primitive.
- **Relay routing bug fix**: `fetch_10002_from_network` now uses
  `fetch_events_from_indexers` instead of `client.fetch_events` (which cannot
  reach DISCOVERY-only indexers — latent bug per the relay taxonomy).
- Concurrent ephemeral relay dedup guard prevents cleanup races between metadata
  and tab-events fetches sharing the same write relays.
- **Follow stats**: pre-populate `CONTACTS_CACHE` from SDK database at login so
  `is_following` is instant on first profile view; fix `AddressViewer` remounting
  `ProfileViewer` when `CLIENT_INITIALIZED` flips (was resetting follower counts
  to 0 after the API response set the correct value).
- **Readiness**: `HAS_SIGNER` gate added to profile effects (canonical pattern);
  live-tail subscription for realtime new posts on the active tab.

### Voice message playback — desktop & Android
- Playback was entirely `#[cfg(feature = "web")]` because the play/pause effect
  used `web_sys` direct DOM access (`get_element_by_id().play()`), compiling out
  on Android (mobile) and desktop — users saw a dead "Playback not supported"
  placeholder.
- Replaced the `web_sys` DOM walk with a uniform `document::eval` JS snippet
  (`getElementById(id).play()/pause()`) that bridges to the real WebView on web,
  desktop (wry), and Android.
- Non-web platforms get a 200ms polling `use_future` reading
  `[currentTime, duration, ended]` via `document::eval`, gated to the active
  player only (at most one eval loop app-wide).
- Card navigation de-gated (mirrors `NoteCard`): plain `use_navigator()`,
  `role="button"`/`tabindex="0"`, unconditional `nav.push(AddressViewer)` on
  click + Enter/Space.

### Media upload refactor
- Splits the overloaded `upload_image()` into purpose-specific paths:
  `upload_image()` handles **images only**, gating compression behind
  `is_image_compressible()` (PNG/JPEG); webp/avif/gif/heic bytes now match the
  declared MIME type.
- New **`upload_file()`** for non-image types (video, audio, PDFs, docs) with no
  compression, replacing the `quality=100` bypass hack.
- `MediaUploader` now accepts configurable `accept` / `max_bytes` per context
  (default `*/*` 10 MB; music raises to 100 MB for audio). All 17 call sites
  updated with explicit accept filters.

### DIFF
--- a/src/components/routstr_settings_panel.rs
+++ b/src/components/routstr_settings_panel.rs
@@ -21,14 +21,14 @@ pub fn RoutstrSettingsPanel(
     let mut balance_error = use_signal(|| None::<String>);
     let mut topup_amount = use_signal(|| "1000".to_string());
     let mut topup_invoice = use_signal(|| None::<RoutstrLightningInvoice>);
-    let topup_loading = use_signal(|| false);
+    let mut topup_loading = use_signal(|| false);
     let mut topup_error = use_signal(|| None::<String>);
     let mut cashu_token_input = use_signal(String::new);
-    let cashu_loading = use_signal(|| false);
+    let mut cashu_loading = use_signal(|| false);
     let mut cashu_error = use_signal(|| None::<String>);
     let mut cashu_success = use_signal(|| None::<String>);
     let mut refund_result = use_signal(|| None::<RoutstrRefundResult>);
-    let refund_loading = use_signal(|| false);
+    let mut refund_loading = use_signal(|| false);
     let mut refund_error = use_signal(|| None::<String>);
     let mut last_loaded_key = use_signal(|| None::<String>);
     let mut request_generation = use_signal(|| 0u32);
@@ -89,10 +89,7 @@ pub fn RoutstrSettingsPanel(
 
     // Invoice payment polling loop (auto-cancels on unmount)
     {
-        let api_key_for_poll = state.read().routstr_api_key.clone();
-        use_future(move || {
-            let api_key_for_poll = api_key_for_poll.clone();
-            async move {
+        use_future(move || async move {
             let mut attempts = 0u32;
             loop {
                 crate::platform::timer::sleep_ms(5000).await;
@@ -114,7 +111,7 @@ pub fn RoutstrSettingsPanel(
                             topup_invoice.set(None);
                             topup_error.set(None);
                             attempts = 0;
-                            if let Some(key) = api_key_for_poll.as_ref() {
+                            if let Some(key) = state.read().routstr_api_key.as_ref() {
                                 if let Ok(result) = routstr::get_balance(key).await {
                                     balance.set(Some(result));
                                 }
@@ -128,7 +125,6 @@ pub fn RoutstrSettingsPanel(
                         }
                     }
             }
-        }
         });
     }
 
@@ -296,6 +292,7 @@ pub fn RoutstrSettingsPanel(
                                     };
                                     let api_key = state.read().routstr_api_key.clone().unwrap_or_default();
                                     topup_error.set(None);
+                                    topup_loading.set(true);
                                     spawn(async move {
                                         match routstr::create_lightning_invoice(amount, "topup", Some(&api_key)).await {
                                             Ok(invoice) => {
@@ -305,6 +302,7 @@ pub fn RoutstrSettingsPanel(
                                                 topup_error.set(Some(err));
                                             }
                                         }
+                                        topup_loading.set(false);
                                     });
                                 },
                                 if *topup_loading.read() { "Creating..." } else { "Create Invoice" }
@@ -380,6 +378,7 @@ pub fn RoutstrSettingsPanel(
                                     return;
                                 }
                                 let api_key = state.read().routstr_api_key.clone().unwrap_or_default();
+                                cashu_loading.set(true);
                                 spawn(async move {
                                     match routstr::topup_with_cashu(&api_key, &token).await {
                                         Ok(result) => {
@@ -401,6 +400,7 @@ pub fn RoutstrSettingsPanel(
                                             cashu_error.set(Some(err));
                                         }
                                     }
+                                    cashu_loading.set(false);
                                 });
                             },
                             if *cashu_loading.read() { "Processing..." } else { "Top Up" }
@@ -419,6 +419,7 @@ pub fn RoutstrSettingsPanel(
                             disabled: *refund_loading.read(),
                             onclick: move |_| {
                                 let api_key = state.read().routstr_api_key.clone().unwrap_or_default();
+                                refund_loading.set(true);
                                 spawn(async move {
                                     match routstr::refund(&api_key).await {
                                         Ok(result) => {
@@ -438,6 +439,7 @@ pub fn RoutstrSettingsPanel(
                                             refund_error.set(Some(err));
                                         }
                                     }
+                                    refund_loading.set(false);
                                 });
                             },
                             if *refund_loading.read() { "Processing..." } else { "Withdraw Remaining Balance" }
@@ -462,11 +464,8 @@ pub fn RoutstrSettingsPanel(
                                             },
                                             "Copy Token"
                                         }
-                                        if let Some(sats) = result.sats {
+                                        if let Some(sats) = result.sats.or_else(|| result.msats.map(|m| m / 1000)) {
                                             span { class: "text-xs text-muted-foreground", "{sats} sats" }
-                                        }
-                                        if let Some(msats) = result.msats {
-                                            span { class: "text-xs text-muted-foreground", "{msats} msats" }
                                         }
                                     }
                                 }
--- a/src/components/routstr_settings_panel.rs
+++ b/src/components/routstr_settings_panel.rs
@@ -1,0 +1,582 @@
+use crate::platform;
+use crate::services::routstr::{
+    self, RoutstrBalance, RoutstrLightningInvoice, RoutstrRefundResult,
+};
+use crate::stores::ai_provider_store::{self, AiProviderState, PROVIDER_STATE_SAVE_EVENT};
+use dioxus::prelude::*;
+use qrcode::{render::svg, QrCode};
+
+#[component]
+pub fn RoutstrSettingsPanel(
+    state: Signal<AiProviderState>,
+    is_saving: Signal<bool>,
+    save_error: Signal<Option<String>>,
+    provider_state_ready: bool,
+) -> Element {
+    let mut api_key_input = use_signal(String::new);
+    let mut key_saving = use_signal(|| false);
+    let mut key_error = use_signal(|| None::<String>);
+    let mut balance = use_signal(|| None::<RoutstrBalance>);
+    let balance_loading = use_signal(|| false);
+    let mut balance_error = use_signal(|| None::<String>);
+    let mut topup_amount = use_signal(|| "1000".to_string());
+    let mut topup_invoice = use_signal(|| None::<RoutstrLightningInvoice>);
+    let topup_loading = use_signal(|| false);
+    let mut topup_error = use_signal(|| None::<String>);
+    let mut cashu_token_input = use_signal(String::new);
+    let cashu_loading = use_signal(|| false);
+    let mut cashu_error = use_signal(|| None::<String>);
+    let mut cashu_success = use_signal(|| None::<String>);
+    let mut refund_result = use_signal(|| None::<RoutstrRefundResult>);
+    let refund_loading = use_signal(|| false);
+    let mut refund_error = use_signal(|| None::<String>);
+    let mut last_loaded_key = use_signal(|| None::<String>);
+    let mut request_generation = use_signal(|| 0u32);
+    let mut pending_save_snapshot = use_signal(|| None::<AiProviderState>);
+    let mut pending_save_min_event_id = use_signal(|| 0u64);
+
+    let current_api_key = state.read().routstr_api_key.clone();
+
+    // Balance loader: re-fetch when api_key changes
+    use_effect(move || {
+        if !provider_state_ready {
+            return;
+        }
+        let key = state.read().routstr_api_key.clone();
+        if last_loaded_key.read().as_ref() == key.as_ref() {
+            return;
+        }
+        last_loaded_key.set(key.clone());
+        balance.set(None);
+        balance_error.set(None);
+        let generation = request_generation.peek().wrapping_add(1);
+        request_generation.set(generation);
+
+        let Some(key) = key else {
+            return;
+        };
+        load_routstr_balance(
+            balance,
+            balance_loading,
+            balance_error,
+            key,
+            Some((request_generation, generation)),
+        );
+    });
+
+    // Save-completion listener (mirrors PPQ panel Effect B)
+    use_effect(move || {
+        let pending_snapshot = pending_save_snapshot.read().clone();
+        let Some(pending_snapshot) = pending_snapshot else {
+            return;
+        };
+        let Some(event) = PROVIDER_STATE_SAVE_EVENT.read().clone() else {
+            return;
+        };
+        if event.event_id <= *pending_save_min_event_id.read()
+            || event.snapshot != pending_snapshot
+        {
+            return;
+        }
+        pending_save_snapshot.set(None);
+        pending_save_min_event_id.set(event.event_id);
+        is_saving.set(false);
+        match event.result {
+            Ok(()) => save_error.set(None),
+            Err(err) => save_error.set(Some(err)),
+        }
+    });
+
+    // Invoice payment polling loop (auto-cancels on unmount)
+    {
+        let api_key_for_poll = state.read().routstr_api_key.clone();
+        use_future(move || {
+            let api_key_for_poll = api_key_for_poll.clone();
+            async move {
+            let mut attempts = 0u32;
+            loop {
+                crate::platform::timer::sleep_ms(5000).await;
+                let invoice = topup_invoice.read().clone();
+                let Some(invoice) = invoice else {
+                    attempts = 0;
+                    continue;
+                };
+                attempts += 1;
+                if attempts > 60 {
+                    topup_error
+                        .set(Some("Payment polling timed out. Please check manually.".to_string()));
+                    topup_invoice.set(None);
+                    attempts = 0;
+                    continue;
+                }
+                if let Ok(status) = routstr::get_lightning_invoice_status(&invoice.invoice_id).await {
+                        if status.status.as_deref() == Some("paid") {
+                            topup_invoice.set(None);
+                            topup_error.set(None);
+                            attempts = 0;
+                            if let Some(key) = api_key_for_poll.as_ref() {
+                                if let Ok(result) = routstr::get_balance(key).await {
+                                    balance.set(Some(result));
+                                }
+                            }
+                        }
+                        if status.status.as_deref() == Some("expired") {
+                            topup_error
+                                .set(Some("Invoice expired. Please try again.".to_string()));
+                            topup_invoice.set(None);
+                            attempts = 0;
+                        }
+                    }
+            }
+        }
+        });
+    }
+
+    let active_key = current_api_key.clone();
+    let active_balance = balance.read().clone();
+    let display_balance_sats = active_balance
+        .as_ref()
+        .and_then(|b| b.balance_msats.map(|m| m / 1000))
+        .unwrap_or(0);
+    let display_total_spent_sats = active_balance
+        .as_ref()
+        .and_then(|b| b.total_spent_msats.map(|m| m / 1000))
+        .unwrap_or(0);
+    let display_total_requests = active_balance
+        .as_ref()
+        .and_then(|b| b.total_requests)
+        .unwrap_or(0);
+
+    let topup_invoice_display = topup_invoice.read().clone();
+    let topup_qr_svg = topup_invoice_display
+        .as_ref()
+        .and_then(|inv| inv.bolt11.as_ref())
+        .map(|b| generate_invoice_qr_svg(b))
+        .unwrap_or_default();
+    let refund_display = refund_result.read().clone();
+
+    rsx! {
+        div { class: "rounded-xl border border-border bg-card p-6 space-y-6",
+            div {
+                h3 { class: "text-lg font-semibold text-foreground", "Routstr" }
+                p { class: "mt-1 text-sm text-muted-foreground",
+                    "Decentralized AI inference marketplace powered by Bitcoin. Pay per request with Lightning or Cashu."
+                }
+            }
+
+            if active_key.is_none() {
+                // API Key entry section
+                div { class: "space-y-3",
+                    label { class: "text-sm font-medium text-foreground", "API Key or Cashu Token" }
+                    input {
+                        r#type: "text",
+                        class: "w-full rounded-lg border border-border bg-background px-3 py-2 text-sm text-foreground placeholder:text-muted-foreground focus:outline-none focus:ring-2 focus:ring-primary",
+                        placeholder: "sk-... or cashuA...",
+                        value: "{api_key_input}",
+                        oninput: move |e| {
+                            api_key_input.set(e.value());
+                            key_error.set(None);
+                        },
+                    }
+                    if let Some(err) = key_error.read().as_ref() {
+                        p { class: "text-xs text-red-600 dark:text-red-400", "{err}" }
+                    }
+                    button {
+                        r#type: "button",
+                        class: "rounded-lg bg-primary px-4 py-2 text-sm font-medium text-primary-foreground transition hover:bg-primary/90 disabled:opacity-60",
+                        disabled: *key_saving.read() || api_key_input.read().trim().is_empty(),
+                        onclick: move |_| {
+                            let input = api_key_input.read().trim().to_string();
+                            if input.is_empty() {
+                                key_error.set(Some("Please enter an API key or Cashu token".to_string()));
+                                return;
+                            }
+                            key_saving.set(true);
+                            key_error.set(None);
+                            let input_clone = input.clone();
+                            spawn(async move {
+                                let final_key = if input_clone.starts_with("cashu") {
+                                    match routstr::create_key_from_cashu(&input_clone).await {
+                                        Ok(key) => key,
+                                        Err(err) => {
+                                            key_error.set(Some(err));
+                                            key_saving.set(false);
+                                            return;
+                                        }
+                                    }
+                                } else {
+                                    input_clone
+                                };
+                                let mut next_state = state.read().clone();
+                                next_state.routstr_api_key = Some(final_key);
+                                persist_routstr_state(
+                                    next_state,
+                                    state,
+                                    is_saving,
+                                    save_error,
+                                    pending_save_snapshot,
+                                    pending_save_min_event_id,
+                                );
+                                api_key_input.set(String::new());
+                                key_saving.set(false);
+                            });
+                        },
+                        if *key_saving.read() { "Saving..." } else { "Save Key" }
+                    }
+                    a {
+                        href: "https://api.routstr.com",
+                        target: "_blank",
+                        class: "block text-xs text-primary hover:underline",
+                        "Create a key at routstr.com →"
+                    }
+                }
+            } else {
+                // Balance + Topup + Refund section
+                div { class: "space-y-4",
+                    // Balance card
+                    div { class: "rounded-lg border border-border p-4",
+                        div { class: "flex items-center justify-between",
+                            div {
+                                p { class: "text-sm text-muted-foreground", "Balance" }
+                                p { class: "text-2xl font-bold text-foreground",
+                                    "{display_balance_sats}"
+                                    span { class: "ml-1 text-sm font-normal text-muted-foreground", "sats" }
+                                }
+                            }
+                            button {
+                                r#type: "button",
+                                class: "rounded-lg border border-border px-3 py-1.5 text-xs transition hover:bg-accent disabled:opacity-60",
+                                disabled: *balance_loading.read(),
+                                onclick: move |_| {
+                                    let key = state.read().routstr_api_key.clone().unwrap_or_default();
+                                    load_routstr_balance(
+                                        balance,
+                                        balance_loading,
+                                        balance_error,
+                                        key,
+                                        None,
+                                    );
+                                },
+                                if *balance_loading.read() { "Loading..." } else { "Refresh" }
+                            }
+                        }
+                        if display_total_requests > 0 || display_total_spent_sats > 0 {
+                            div { class: "mt-2 flex gap-4 text-xs text-muted-foreground",
+                                span { "{display_total_requests} requests" }
+                                span { "{display_total_spent_sats} sats spent" }
+                            }
+                        }
+                        if let Some(err) = balance_error.read().as_ref() {
+                            p { class: "mt-2 text-xs text-red-600 dark:text-red-400", "{err}" }
+                        }
+                    }
+
+                    // Top up via Lightning
+                    div { class: "rounded-lg border border-border p-4 space-y-3",
+                        p { class: "text-sm font-medium text-foreground", "Top Up via Lightning" }
+                        div { class: "flex gap-2",
+                            input {
+                                r#type: "number",
+                                class: "w-32 rounded-lg border border-border bg-background px-3 py-2 text-sm text-foreground focus:outline-none focus:ring-2 focus:ring-primary",
+                                placeholder: "sats",
+                                value: "{topup_amount}",
+                                oninput: move |e| { topup_amount.set(e.value()); },
+                            }
+                            button {
+                                r#type: "button",
+                                class: "rounded-lg bg-primary px-4 py-2 text-sm font-medium text-primary-foreground transition hover:bg-primary/90 disabled:opacity-60",
+                                disabled: *topup_loading.read(),
+                                onclick: move |_| {
+                                    let amount = match topup_amount.read().trim().parse::<u64>() {
+                                        Ok(a) if a > 0 => a,
+                                        _ => {
+                                            topup_error.set(Some("Enter a valid amount in sats".to_string()));
+                                            return;
+                                        }
+                                    };
+                                    let api_key = state.read().routstr_api_key.clone().unwrap_or_default();
+                                    topup_error.set(None);
+                                    spawn(async move {
+                                        match routstr::create_lightning_invoice(amount, "topup", Some(&api_key)).await {
+                                            Ok(invoice) => {
+                                                topup_invoice.set(Some(invoice));
+                                            }
+                                            Err(err) => {
+                                                topup_error.set(Some(err));
+                                            }
+                                        }
+                                    });
+                                },
+                                if *topup_loading.read() { "Creating..." } else { "Create Invoice" }
+                            }
+                        }
+                        if let Some(err) = topup_error.read().as_ref() {
+                            p { class: "text-xs text-red-600 dark:text-red-400", "{err}" }
+                        }
+                        if let Some(invoice) = topup_invoice_display.as_ref() {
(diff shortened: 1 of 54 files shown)


goodmorningbitcoin (commented):
Review batch 1: New files (routstr service + settings panel)

src/components/routstr_settings_panel.rs
```diff
@@ -0,0 +1,582 @@
...
+        is_saving.set(false);
+        match event.result {
+            Ok(()) => save_error.set(None),
+            Err(err) => save_error.set(Some(err)),
+        }
+    });
+
+    // Invoice payment polling loop (auto-cancels on unmount)
+    {
+        let api_key_for_poll = state.read().routstr_api_key.clone();
```
goodmorningbitcoin: ⚠️ **Stale API key in payment poll loop**

`api_key_for_poll` is read from `state` once at component mount and captured into the `use_future` closure. If the user replaces their API key (via "Replace Key") while a topup invoice is pending, the post-payment balance refresh (line ~117) will use the *old* key — `routstr::get_balance(key)` returns a 401 and the balance silently doesn't update.

**Fix:** Read the key from `state` inside the loop body (like the `topup_invoice.read()` call already does):
```rust
// After detecting "paid":
let key = state.read().routstr_api_key.clone();
if let Some(key) = key {
    if let Ok(result) = routstr::get_balance(&key).await {
        balance.set(Some(result));
    }
}
```

src/services/routstr.rs
```diff
@@ -0,0 +1,311 @@
...
+    pub msats: Option<u64>,
+    pub raw_json: String,
+}
+
+pub async fn get_balance(api_key: &str) -> Result<RoutstrBalance, String> {
+    let url = format!("{}/balance/info", ROUTSTR_BASE_URL);
+    let headers = vec![("Authorization".to_string(), format!("Bearer {}", api_key))];
+    let value = send_request(Method::GET, &url, Some(headers), None).await?;
+    Ok(RoutstrBalance {
+        balance_msats: number_field(&value, &["balance"]).map(|n| n as u64),
```
goodmorningbitcoin, line 55: ⚠️ **f64→u64 truncation risk for financial amounts**

`number_field` returns `f64`, which is then cast `as u64`. JSON numbers are parsed as `f64` by `serde_json`. For values above 2^53 (~9 PB of msats), this loses integer precision. While unlikely for a single user's balance, the `total_spent_msats` field is cumulative and could theoretically grow large. More practically, if the API returns a value like `1e20` (scientific notation), `as u64` produces a wrapped value rather than an error.

**Fix:** Parse the raw JSON value as `u64` first, falling back to `f64` only for string-encoded numbers:
```rust
fn number_field_u64(value: &Value, keys: &[&str]) -> Option<u64> {
    for key in keys {
        if let Some(found) = value.get(*key) {
            if let Some(n) = found.as_u64() { return Some(n); }
            if let Some(s) = found.as_str() { return s.parse::<u64>().ok(); }
            if let Some(n) = found.as_f64() { return Some(n as u64); }
        }
    }
    None
}
```

src/components/routstr_settings_panel.rs
```diff
@@ -0,0 +1,582 @@
...
+                                            class: "rounded-lg border border-border px-3 py-1.5 text-xs transition hover:bg-accent",
+                                            onclick: move |_| {
+                                                let token = refund_result.read().as_ref().and_then(|r| r.token.clone()).unwrap_or_default();
+                                                spawn(async move {
+                                                    let _ = platform::clipboard::copy_to_clipboard(&token).await;
+                                                });
+                                            },
+                                            "Copy Token"
+                                        }
+                                        if let Some(sats) = result.sats {
```
goodmorningbitcoin: 💡 **Nitpick: Refund display shows both sats and msats simultaneously**

The refund result renders both `{sats} sats` and `{msats} msats` when both fields are present. Since the Routstr refund endpoint returns both fields for the same amount, the user sees something like `1000 sats 1000000 msats` — confusing since they represent the same balance.

**Fix:** Show only sats (converting from msats if sats is None), or show msats only when sats is None.

src/stores/nostr_client/streaming.rs
```diff
@@ -299,6 +299,173 @@ where
...
+            }
+        });
+    }
+
+    drop(tx); // All senders spawned; drop original so rx ends when all finish.
+
+    // ── Consumer: merge with EventId dedup + author filter ──
+    let mut seen: std::collections::HashSet<nostr::EventId> = std::collections::HashSet::new();
+    let mut count = 0;
+    while let Some(event) = rx.recv().await {
```
goodmorningbitcoin, line 454: ⚠️ **No overall deadline on the consumer merge loop**

The consumer loop `while let Some(event) = rx.recv().await` blocks until all three spawned senders drop their `tx` clones. Each sender's `stream_events_from`/`stream_events` call has its own `timeout`, but the consumer has no independent overall deadline. If one spawned task hangs beyond its timeout — say due to a tokio scheduling stall or a relay that holds the connection open without sending EOSE — the consumer blocks until that task's stream resolves.

In practice the per-call `timeout` bounds each source, so the consumer wakes when the last source completes. But consider adding a `tokio::time::timeout(timeout * 2, ...)` wrapper as defense-in-depth so a stuck sender can't hold the consumer hostage indefinitely.

src/stores/mostro/cleanup.rs
```diff
@@ -352,33 +370,33 @@ mod tests {
...
         t.order_id = "maker-5".to_string();
-        assert!(!is_orphan(&t, NOW), "young placeholder maker is not orphan");
+        assert!(!is_orphan(&t, NOW, false), "young placeholder maker is not orphan");
     }
 
     #[test]
     fn test_orphan_within_bond_grace_is_kept() {
         let mut t = build_trade(TradeStatus::Pending, TradeRole::Taker, 150, false);
         t.is_bond_invoice = Some(true);
         assert!(
```
goodmorningbitcoin, line 380: 💡 **Suggestion: Bond grace test doesn't distinguish bond grace from regular threshold**

`test_orphan_within_bond_grace_is_kept` uses `age=150` with `is_bond_invoice=true`. With the new thresholds (`ORPHAN_THRESHOLD_SECS=600`, `ORPHAN_BOND_GRACE_SECS=900`), age 150 is under *both* thresholds, so this test would pass even if the bond grace logic were removed entirely.

**Fix:** Use `age=650` (> 600 regular threshold, < 900 bond grace) so the test actually exercises the bond grace path:
```rust
let mut t = build_trade(TradeStatus::Pending, TradeRole::Taker, 650, false);
t.is_bond_invoice = Some(true);
assert!(!is_orphan(&t, NOW, false), "bond-flow trade within grace window is not orphan");
```

src/components/viewers/profile_viewer.rs
```diff
@@ -516,118 +602,188 @@ pub fn ProfileViewer(pubkey: String) -> Element {
...
-                        filter, Duration::from_secs(5),
-                    );
+                    // Stream events from the author's outbox relays for
+                    // progressive UI updates — posts paint as they arrive
+                    // instead of blocking for the full EOSE window. Mirrors
+                    // the home Following feed's `load_following_feed_streaming`
+                    // pattern: per-event callback + `DebouncedCollector` for
+                    // coalesced signal writes.
+                    let collector =
+                        crate::utils::debounced_collector::DebouncedCollector::<NostrEvent>::new(
```
goodmorningbitcoin, line 612: ⚠️ **Merge/dedup logic duplication between streaming callback and tail drain**

The event merge logic (dedup against existing IDs, sort by created_at/published_at, compute oldest_timestamp, update tab_data) is duplicated verbatim between the streaming `collector.extend` callback (~lines 638-690) and the tail `collector.drain()` block (~lines 713-760). Any future bug fix or behavioral change needs to be applied in both places, and the two copies have already drifted slightly in variable naming.

**Fix:** Extract a helper function `merge_tab_events(td, tab, new_events, pc)` and call it from both the callback and the drain.

src/routes/mostro/mostro_order_detail.rs
```diff
@@ -114,6 +114,12 @@ fn render_order_detail(order: &P2POrder, mut recovering: Signal<bool>) -> Elemen
     // Bound before the rsx so the recover button's 'static onclick can
     // capture an owned order_id (the borrowed `order` can't escape).
     let recover_oid = order.order_id.clone();
+    // Check if the current user owns this order (as maker or taker) via
+    // the local TRADES cache or the durable creation_ledger.
```
goodmorningbitcoin, line 118: ⚠️ **`is_owned` check is too broad — shows "Open Trade" for any ledger match**

`is_owned` checks if the order exists in `TRADES` *or* `creation_ledger::entries_for_order`. But `entries_for_order` returns entries for *any* role (Maker or Taker). This means a user who *took* someone else's order will see "You own this order" + "Open Trade" badge on the public order detail page, which is misleading — they don't "own" the listing, they have a trade position on it.

Consider filtering by Maker role if the intent is to show ownership of the listing:
```rust
let is_owned = crate::stores::mostro::find_by_order_id(&owned_order_id)
    .is_some_and(|t| t.role == TradeRole::Maker)
    || crate::stores::mostro::creation_ledger::entries_for_order(&owned_order_id)
        .iter().any(|e| e.role == TradeRole::Maker);
```

src/components/voice/message_card.rs
```diff
@@ -319,6 +289,48 @@ pub fn VoiceMessageCard(
         voice_messages_store::pause_voice_message();
         current_time.set(0.0);
     };
+    #[cfg(not(feature = "web"))]
+    {
+        let audio_id_poll = audio_id.clone();
+        use_future(move || {
```
goodmorningbitcoin, line 295: ⚠️ **Per-card polling `use_future` on non-web platforms**

The `use_future` polls every 200ms via `document::eval` on desktop/mobile. The loop correctly skips the JS eval when the card isn't the active player (`!= Some(event_id)`), but every `VoiceMessageCard` instance still wakes up, reads the signal, and loops every 200ms. In a feed with 20 voice messages, that's 20 `use_future` tasks waking up every 200ms.

Consider collapsing to a single app-wide poller (like the web `_handle_timeupdate` event-based pattern) or having the component return early from the `use_future` body when it's not the active player, rather than spinning the loop.

goodmorningbitcoin (commented):
Review batch 2: Modified files (streaming, cleanup, profile viewer, mostro, voice)

goodmorningbitcoin:
## Code Review Summary

**Verdict: Comment** — 0 blockers, 6 warnings, 2 suggestions

Reviewed 48 files (+2,791/-680) across five feature areas. Overall this is a well-structured release — the Routstr provider integration, Mostro trade durability hardening, and profile streaming optimization are architecturally sound. No critical/security issues found.

### ⚠️ Warnings

- **`streaming.rs:454`** — Consumer merge loop in `stream_profile_events_from_relays` has no overall deadline; relies on per-source timeouts
- **`routstr_settings_panel.rs:92`** — Payment poll loop captures API key at mount; stale key used after "Replace Key" 
- **`routstr.rs:55`** — `number_field` uses `f64`→`u64` cast; precision risk for large financial amounts
- **`profile_viewer.rs:612`** — Merge/dedup logic duplicated verbatim between streaming callback and tail drain (~50 lines each)
- **`mostro_order_detail.rs:118`** — `is_owned` shows "Open Trade" for any creation_ledger match regardless of role (Maker vs Taker)
- **`message_card.rs:295`** — Non-web polling `use_future` runs per-card; N voice cards = N 200ms pollers

### 💡 Suggestions / Nitpicks

- **`cleanup.rs:380`** — Bond grace test uses age=150 which passes both thresholds; should use 650 to test bond grace specifically
- **`routstr_settings_panel.rs:465`** — Refund display shows both sats AND msats for the same amount

### ✅ Looks Good

- **Routstr integration**: Defensive JSON parsing with redacted error bodies, proper generation-based stale-task prevention, NIP-44 encrypted credential sync
- **Mostro durability**: Placeholder-before-send pattern correctly cleans up on send failure AND daemon rejection; `RecvError::Lagged` correctly treated as recoverable
- **Profile streaming**: Multi-source merge with EventId dedup mirrors SDK pool pattern; ephemeral relay dedup guard prevents cleanup races
- **Media upload refactor**: Clean separation of `upload_image` (compressible) vs `upload_file` (passthrough); `is_image_compressible` correctly gates UI quality slider
- **Mention autocomplete fix**: Correct diagnosis of `get_bounding_client_rect()` returning viewport-relative coords
- **Voice playback**: `serde_json::to_string` for element ID is safe against injection in `document::eval`
- **Test coverage**: New unit tests for routstr parsing, cleanup orphan thresholds, and provider setup gates

---
*Reviewed by Hermes Agent*

src/components/routstr_settings_panel.rs
```diff
@@ -0,0 +1,582 @@
...
+) -> Element {
+    let mut api_key_input = use_signal(String::new);
+    let mut key_saving = use_signal(|| false);
+    let mut key_error = use_signal(|| None::<String>);
+    let mut balance = use_signal(|| None::<RoutstrBalance>);
+    let balance_loading = use_signal(|| false);
+    let mut balance_error = use_signal(|| None::<String>);
+    let mut topup_amount = use_signal(|| "1000".to_string());
+    let mut topup_invoice = use_signal(|| None::<RoutstrLightningInvoice>);
+    let topup_loading = use_signal(|| false);
```
goodmorningbitcoin: 🔴 **Critical: `topup_loading`, `cashu_loading`, and `refund_loading` are never set to `true`**

All three signals are declared as `use_signal(|| false)` (lines 24, 27, 31) and used for button `disabled` / text, but they are **never set to `true`** inside their respective onclick handlers. The async operations run without any UI feedback or submission guard.

This means:
- **Duplicate submissions**: User can click "Create Invoice" multiple times, creating multiple Lightning invoices
- **Duplicate Cashu topups**: User can submit the same token twice
- **Duplicate refunds**: User can click "Withdraw" multiple times — the second call may refund an already-zero balance

The refund case is the most dangerous — a double-submit could produce confusing error states or race with the balance refresh.

**Fix:** Set `*_loading.set(true)` before the `spawn` and `.set(false)` after the async work completes.

src/services/routstr.rs
```diff
@@ -0,0 +1,311 @@
...
+    Ok(RoutstrTopupResult {
+        msats: number_field(&value, &["msats"]).map(|n| n as u64),
+        raw_json: pretty_json(&value),
+    })
+}
+
+pub async fn create_key_from_cashu(cashu_token: &str) -> Result<String, String> {
+    let base = format!("{}/balance/create", ROUTSTR_BASE_URL);
+    let mut url = Url::parse(&base).map_err(|e| format!("Invalid Routstr URL: {}", e))?;
+    url.query_pairs_mut()
```
goodmorningbitcoin, line 123: 🔴 **Critical: Cashu token passed as GET query parameter**

`create_key_from_cashu` puts the Cashu token (`cashuA...`) in the URL query string:
```rust
url.query_pairs_mut()
    .append_pair("initial_balance_token", cashu_token);
```

Cashu tokens are bearer instruments with monetary value. GET query parameters appear in:
- Server access logs (Routstr API server, CDN, reverse proxy)
- Browser history (on WASM)
- HTTP `Referer` headers if the page navigates after the request

**Fix:** If the Routstr API supports it, use a POST body instead. If the API only accepts GET, document this as a known exposure.

src/stores/relay/coverage.rs
```diff
@@ -335,10 +357,13 @@ pub async fn connect_ephemeral_relays(client: &Client, urls: &[String]) -> Ephem
         let Ok(parsed) = nostr::Url::parse(&url) else {
             continue;
         };
+        // Check if the relay is already connected in the pool.
         let is_connected = relays
             .iter()
             .any(|(u, r)| u.as_str() == url && r.is_connected());
-        if is_connected {
+        // Check if another concurrent caller has claimed this relay.
+        let in_use = EPHEMERAL_IN_USE.lock().unwrap().contains(&url);
```
goodmorningbitcoin: 🔴 **TOCTOU race in `connect_ephemeral_relays` dedup**

The `EPHEMERAL_IN_USE` check (line 365) and the subsequent claim (lines 381-385) are two separate lock acquisitions. Between them, `client.relays().await` at line 351 yields to the executor, allowing two concurrent callers to both see `in_use = false`, both add the URL to `to_add`, and both claim it. The dedup is defeated.

**Fix:** Check-and-claim atomically in a single lock scope — check `EPHEMERAL_IN_USE` and insert immediately if absent, all under one `lock()` call. Move relay pool queries outside the critical section.

goodmorningbitcoin (commented):
Review batch 3: Deep-dive findings — loading states, Cashu token exposure, relay coverage races

src/stores/relay/coverage.rs
```diff
@@ -391,12 +424,21 @@ pub async fn connect_ephemeral_relays(client: &Client, urls: &[String]) -> Ephem
 }
```
goodmorningbitcoin, line 439: ⚠️ **Failed ephemeral relay connections leak `EPHEMERAL_IN_USE` entries**

`connect_ephemeral_relays` claims URLs in `EPHEMERAL_IN_USE` (lines 381-385) before connecting. But if `try_connect_relay` fails (returns `Err`), the URL is NOT included in `newly_connected` and thus never reaches `cleanup_ephemeral_relays` — which is the only place that removes entries from `EPHEMERAL_IN_USE`. A transiently-unreachable relay URL is permanently stuck, blocking all future reconnection attempts by other callers.

**Fix:** After the `buffer_unordered` collect, remove failed URLs from `EPHEMERAL_IN_USE`:
```rust
let mut in_use = EPHEMERAL_IN_USE.lock().unwrap();
for url in &to_add {
    if !newly_connected.contains(url) {
        in_use.remove(url);
    }
}
```

goodmorningbitcoin (commented):
Batch 4-0

src/stores/mostro/cleanup.rs
```diff
@@ -56,9 +57,9 @@ fn effective_max_age_secs() -> u64 {
 }
 
 /// A Pending trade with no daemon reply older than this is an orphan.
```
goodmorningbitcoin, line 59: ⚠️ **Module doc comments still reference old threshold values**

The module-level doc at lines 22-24 says "default 120s" and lines 31-32 say "default 180s", but the constants were changed to `ORPHAN_THRESHOLD_SECS: i64 = 600` and `ORPHAN_BOND_GRACE_SECS: i64 = 900`. The documentation now contradicts the code.

**Fix:** Update the doc comments to reflect the new values (600s / 900s).

goodmorningbitcoin (commented):
Batch 4-1 (fixed line)

src/stores/mostro/node_config.rs
```diff
@@ -270,9 +270,9 @@ impl MostroNodeInfo {
                 info.hold_invoice_cltv_delta = val.parse().ok();
             } else if kind == TagKind::Custom(std::borrow::Cow::Borrowed("invoice_expiration_window")) {
                 info.invoice_expiration_window = val.parse().ok();
-            } else if kind == TagKind::Custom(std::borrow::Cow::Borrowed("lnd_node_uri")) {
+            } else if kind == TagKind::Custom(std::borrow::Cow::Borrowed("lnd_uris")) {
```
goodmorningbitcoin: ⚠️ **`lnd_uris` tag rename lacks backward-compat fallback**

The tag name changed from `"lnd_node_uri"` to `"lnd_uris"` but the old tag is not kept as a fallback. Any deployed daemon still publishing the `lnd_node_uri` tag will have this field silently dropped.

**Fix:** Parse both tag names — try `"lnd_uris"` first, fall back to `"lnd_node_uri"`:
```rust
} else if kind == TagKind::Custom(Cow::Borrowed("lnd_uris"))
    || kind == TagKind::Custom(Cow::Borrowed("lnd_node_uri"))
{
    info.lnd_uris = Some(val);
}
```

goodmorningbitcoin (commented):
Batch 4-2

src/stores/nostr_client/contacts.rs
```diff
@@ -446,3 +445,34 @@ pub async fn follow_users_batch(
...
+/// Pre-populate CONTACTS_CACHE from the SDK database (no network).
+///
+/// Called at login (`warmup_profiles_from_db`) so that `is_following` is
+/// instant on the first profile view of a session. Without this, the first
+/// `fetch_contacts` call for the current user blocks up to 5s on a relay
+/// round-trip, leaving the Follow button in the wrong state.
+///
+/// Only stores pubkeys (no relay_url/petname) — sufficient for `is_following`.
+/// The full enriched contacts are fetched later by `warmup_profiles_from_network`.
+pub fn warm_contacts_cache(pubkey: &str, contact_pubkeys: Vec<String>) {
```
goodmorningbitcoin, line 458: ⚠️ **`warm_contacts_cache` unconditionally overwrites richer existing data**

If `warmup_profiles_from_db` runs after a prior `fetch_contacts` already populated the cache with enriched contacts (including `relay_url` and `petname`), `warm_contacts_cache` replaces it with pubkey-only `EnrichedContact::new` entries — losing the relay_url/petname data until the next network refresh.

This is an edge case (warmup runs at login, before network fetch), but on re-login without full logout the cache may survive.

**Fix:** Guard the overwrite:
```rust
if cache.as_ref().map(|c| c.pubkey == pubkey && !c.contacts.is_empty()).unwrap_or(false) {
    return; // Don't clobber a richer existing cache
}
```

goodmorningbitcoin (commented):
Batch 4-3 (fixed line)

goodmorningbitcoin:
## Code Review Summary — Updated

**Verdict: Comment** — 3 critical, 9 warnings, 5 suggestions/nitpicks

Reviewed 48 files (+2,791/-680) across five feature areas using parallel deep-dive analysis. Overall architecture is sound — the Routstr integration, Mostro durability, and profile streaming are well-structured. No blockers, but 3 critical issues should be addressed before merge.

### 🔴 Critical (3)

- **`routstr_settings_panel.rs:24`** — `topup_loading`, `cashu_loading`, `refund_loading` signals never set to `true`. All three async operations allow duplicate submissions — especially dangerous for refund/withdraw.
- **`routstr.rs:117`** — Cashu token (bearer instrument with monetary value) passed as GET query parameter. Appears in server/proxy/CDN logs. Should be POST body if API supports it.
- **`coverage.rs:365`** — TOCTOU race in `EPHEMERAL_IN_USE` check-and-claim. Two concurrent callers can both see `in_use = false` and both claim the same relay URL.

### ⚠️ Warnings (9)

- **`streaming.rs:454`** — Consumer merge loop has no overall deadline; relies on per-source timeouts
- **`routstr_settings_panel.rs:92`** — Payment poll captures API key at mount; stale key used after "Replace Key"
- **`routstr.rs:55`** — `number_field` uses `f64`→`u64` cast; precision risk for financial amounts
- **`coverage.rs:424`** — Failed ephemeral relay connections leak `EPHEMERAL_IN_USE` entries permanently
- **`cleanup.rs:59`** — Module doc comments still say "default 120s/180s" but constants are 600/900
- **`node_config.rs:273`** — `lnd_uris` tag rename lacks backward-compat fallback for older daemons
- **`contacts.rs:458`** — `warm_contacts_cache` unconditionally overwrites richer existing cache data
- **`profile_viewer.rs:612`** — Merge/dedup logic triplicated (streaming callback, tail drain, live-tail)
- **`mostro_order_detail.rs:118`** — `is_owned` shows "Open Trade" for any creation_ledger match regardless of role

### 💡 Suggestions / Nitpicks (5)

- **`cleanup.rs:380`** — Bond grace test uses age=150 which passes both thresholds; should use 650
- **`routstr_settings_panel.rs:465`** — Refund display shows both sats AND msats for same amount
- **`message_card.rs:295`** — Per-card polling on non-web; N voice cards = N pollers
- **`blossom_store.rs:303`** — Doc comment placement error merges `is_image_compressible` docs into `upload_image`
- **`restore.rs`** — `recover_order_by_id` listener treats all errors as fatal (no Lagged recovery)

### ✅ Looks Good

- Routstr: defensive JSON parsing with redacted error bodies, NIP-44 encrypted credential sync, good test coverage
- Mostro: placeholder-before-send pattern with correct cleanup on send failure + daemon rejection
- Profile streaming: multi-source merge with EventId dedup, ephemeral relay dedup guard concept
- Media upload: clean `upload_image` vs `upload_file` separation, correct compression gating
- Mention autocomplete: correct `getBoundingClientRect` → viewport-relative fix
- take.rs: `sats_from_override` → `range_fiat_amount` is a correct bug fix (was sending sats where fiat expected)
- Voice: `serde_json::to_string` for element ID is safe against injection

---
*Reviewed by Hermes Agent — 15 inline comments across 5 review batches*

src/components/routstr_settings_panel.rs
patrickulrich: **Fixed.** All three signals are now declared `mut` and follow the PPQ panel pattern: each onclick sets `*_loading.set(true)` after validation passes (before `spawn`) and `*_loading.set(false)` at the end of the spawned async body. This prevents duplicate submissions on all three operations.

src/stores/relay/coverage.rs
patrickulrich: **Fixed.** The check and claim are now merged into a single `EPHEMERAL_IN_USE.lock()` scope using `HashSet::insert()` return value for atomic check-and-claim. `insert()` returns `false` if already present (claimed by another caller), so check and claim happen in one operation — no TOCTOU window. The lock guard is dropped before any `.await`, preserving the `StdMutex` convention. This was strictly required for desktop (multi-threaded tokio); WASM was accidentally safe due to single-threaded cooperative scheduling.

src/stores/relay/coverage.rs
patrickulrich: **Fixed.** After the `buffer_unordered` collect, we now compute `failed = claimed_urls − newly_connected` and for each failed URL: `EPHEMERAL_IN_USE.lock().remove(url)` + `client.force_remove_relay(url).await`. This prevents permanently stuck entries that would silently break all future connection attempts to that relay URL.

src/components/routstr_settings_panel.rs
patrickulrich: **Fixed.** Removed the mount-time `api_key_for_poll` capture entirely. The balance refresh now reads the key fresh inside the loop body via `state.read().routstr_api_key.as_ref()`, mirroring how `topup_invoice.read()` is already used at line 99. Verified via Dioxus source that `use_future` does NOT establish a `ReactiveContext` (unlike `use_effect`/`use_resource`), so reading a Signal inside the async body is safe and will not create a reactive dependency that restarts the future.

src/stores/mostro/cleanup.rs
patrickulrich: **Fixed.** Module doc comments updated: `120s → 600s` and `180s → 900s` to match the actual constants (`ORPHAN_THRESHOLD_SECS = 600`, `ORPHAN_BOND_GRACE_SECS = 900`).

src/stores/mostro/node_config.rs
patrickulrich: **Fixed.** Added backward-compat fallback: the tag parser now accepts both `"lnd_uris"` (spec-conformant) and `"lnd_node_uri"` (legacy) tag names, mapping both to `info.lnd_uris`. This app is a consumer-only (daemons publish the tag), so supporting the old name ensures compatibility with deployed daemons still emitting `lnd_node_uri`.

src/stores/mostro/cleanup.rs
patrickulrich: **Fixed.** Changed test age from 150 to 650. With 650: `650 > 600` (regular threshold → would be orphan) but `650 < 900` (bond grace → not orphan), so the test now specifically exercises the bond grace path. The companion test `test_orphan_past_bond_grace_is_removed` already used 950 (> 900), so the pair (650 kept, 950 removed) correctly brackets the 900 boundary.

src/routes/mostro/mostro_order_detail.rs
patrickulrich: **Fixed.** The `is_owned` check now also determines `is_maker` (following the pattern in `take_mostro_button.rs:23-27`) by checking `TradeRole::Maker` on both the TRADES cache entry and creation_ledger entries. The label is now role-aware: Makers see "You own this order", Takers see "You have a trade on this order". The "Open Trade →" shortcut remains visible for both roles since Takers also need to manage their trades.

src/components/routstr_settings_panel.rs
patrickulrich: **Fixed.** Replaced the two separate `if let Some(sats)` / `if let Some(msats)` spans with a single display: `result.sats.or_else(|| result.msats.map(|m| m / 1000))` — shows sats only, deriving from msats if sats is absent.

src/stores/nostr_client/contacts.rs
patrickulrich: **Fixed.** Added a guard at the top of `warm_contacts_cache`: if the cache already holds enriched contacts (with `relay_url` or `petname`) for the same pubkey, the warmup is skipped. This preserves richer data on re-login without full logout. The check acquires the cache lock, inspects for enriched entries, and returns early — only pubkey-only caches or different-user caches proceed with the warmup.

src/components/voice/message_card.rs
patrickulrich: **Valid concern — deferring (architecture change).** The current loop correctly skips the JS `eval` when the card is not the active player (`!= Some(event_id)`), so the cost per inactive card is just a Signal read + a 200ms sleep, not a JS eval. A single app-wide poller (like the web `_handle_timeupdate` event-based pattern) would be ideal but requires a shared mutable "active player state" singleton with cross-component coordination. This is a larger architecture change best done as a dedicated follow-up. The practical impact is bounded since feeds rarely show more than a few voice messages simultaneously.

src/stores/nostr_client/streaming.rs
patrickulrich: **Valid defense-in-depth suggestion — deferring.** Each of the three spawned senders is bounded by the per-call `timeout` (8s), so the consumer loop wakes when the last source completes. Adding a `timeout(timeout * 2, ...)` wrapper as suggested would be a clean safety net, but the per-source timeouts already prevent indefinite hangs in practice. Will consider for a follow-up hardening pass.

src/components/viewers/profile_viewer.rs
patrickulrich: **Valid code quality concern — deferring to avoid regression risk.** The merge/dedup logic is ~50 lines triplicated between the streaming callback, tail drain, and (conceptually) the live-tail handler. Extracting a `merge_tab_events(td, tab, new_events, pc)` helper is the right cleanup. However, this PR already touches 48 files; a refactor of working streaming code carries regression risk disproportionate to the benefit. Will do as a focused follow-up PR.

src/services/routstr.rs
patrickulrich: **Valid but low practical risk — leaving as-is.** The `f64` cast precision loss starts at 2^53 msats (~9 petabytes of sats). For a single user's balance or even cumulative `total_spent_msats`, this is astronomically unlikely. All other Routstr endpoints return similar small msats integers. The existing `number_field` also handles string-encoded numbers (some endpoints return `"sats": "1000"`), so switching to a u64-first parser would add complexity for negligible gain. Happy to revisit if the API ever returns values in the 2^53 range.

src/services/routstr.rs
patrickulrich: **Acknowledged — API limitation, cannot fix client-side.** I verified the routstr-core server source (`routstr/balance.py`): the `/v1/balance/create` endpoint is defined as `@router.get` — GET only. The routstr-core codebase has a TODO to implement a POST `/v1/wallet/create` endpoint but it does not exist yet. All other endpoints that accept Cashu tokens (`/topup`, `/refund`) already use POST with JSON bodies in our client. When routstr-core ships the POST endpoint, we will switch immediately.

PR #352, opened 2026-08-06
Release 0.8.21: Deflock + thread-scroll and relay race bug fixes

### LINKED ISSUE #348: Bug: Clicking a thread opens scrolled down instead of at the top
## Bug Description

When clicking a note/thread from the feed, the page opens scrolled down partway instead of starting at the top. The user is "thrown down the page" and has to scroll back up to see the original post.

## Steps to Reproduce

1. Open the home feed (scroll down a few screens so there's scroll position)
2. Click any note in the feed to open the thread detail view
3. Observe the page loads scrolled down — you see replies or mid-thread content instead of the original post at the top

## Expected Behavior

The thread detail page should open scrolled to the top (y=0), showing the original post first, then parents (if any), then replies below.

## Actual Behavior

The page opens at a non-zero scroll position. The user sees content from the middle of the thread.

## Root Cause

The bug is a race condition between scroll restoration and async content loading in the `AddressViewer` → `NoteViewer` pipeline.

### The navigation flow:

1. **`NoteCard::onclick`** (`src/components/note_card.rs:317-333`) calls `nav.push(Route::AddressViewer { address })`.

2. **Dioxus `WebHistory::push()`** (`dioxus-web-0.7.5/src/history.rs:136-151`) saves the current scroll position to history state via `update_scroll()`, then calls `handle_nav()` which immediately calls `window.scroll_to_with_x_and_y(0.0, 0.0)`.

3. **`Layout`'s route-change `use_effect`** (`src/routes/mod.rs:987-1011`) fires. It checks `was_popstate_nav()` → returns `false` for a forward push → calls `set_scroll_y(0.0)`. This spawns an async `document::eval("window.scrollTo(0, 0)")`.

4. **`AddressViewer`** (`src/routes/address_viewer.rs:11-256`) renders `AddressState::Loading` — a centered spinner with `min-h-screen`. The page has almost no content. The scroll-to-zero from step 3 executes against this near-empty page.

5. **`resolve_address()`** completes asynchronously, transitioning state to `AddressState::Note { ... }`.

6. **`NoteViewer`** (`src/components/viewers/note_viewer.rs:870-1413`) mounts and begins its own async data loading. The `use_effect` at line 917 fetches the main note, parents, and replies — all of which grow the page height substantially.

7. **No scroll reset happens after content loads.** The `set_scroll_y(0.0)` from step 3 already fired against the spinner page. When the actual thread content renders (parents + main note + replies), the browser layout recalculates and the page height grows — but the scroll position is not corrected.

### Why the user gets "thrown down":

- The `window.scrollTo(0, 0)` call in step 2/3 runs when `AddressViewer` is showing a `min-h-screen` spinner. At that moment, `document.body.scrollHeight` is approximately `window.innerHeight` — so `scrollY` is already 0.
- When `NoteViewer` renders parent events above the main note (the `border-b-2 border-blue-500/20` parent chain at line 1289), the page grows upward/downward. Parent notes are prepended, pushing the main note down.
- If any async reply data arrives and the thread tree rebuilds (the `thread_tree_memo` at line 903), the page height changes again.
- The browser's scroll position is not recalculated to account for the new content. The user sees wherever the layout happened to settle.

### Additional contributing factor: `nav.replace()` in `Note` route

The `Note` route (`src/routes/note.rs:5-9`) immediately calls `nav.replace(Route::AddressViewer { address: note_id })`. This `replace` triggers `WebHistory::replace()` which calls `handle_nav()` (scrolls to 0) — but again, this fires before any content has rendered.

### Missing: No scroll-to-top after content load

`NoteViewer`'s `use_effect` (line 917) handles data loading but never calls `set_scroll_y(0.0)` after `loading.set(false)` or after `note_data` is populated. There is no post-render scroll correction.

### Fix 1: Scroll to top after NoteViewer content loads (Primary fix)

Add a scroll-to-top call in `NoteViewer`'s `use_effect` after the main note data is loaded and the component is about to render the actual content. The best insertion point is right after `loading.set(false)` on line 1000:

```rust
// src/components/viewers/note_viewer.rs, inside the use_effect spawn block
// after note_data.set(Some(event.clone())) and loading.set(false)
note_data.set(Some(event.clone()));
loading.set(false);
// ... parent_ids extraction ...

// Scroll to top after content is ready
spawn(async move {
    crate::stores::ui::scroll_restore::set_scroll_y(0.0).await;
});
```

And/or add a separate `use_effect` that watches `note_data` and fires a scroll reset when it transitions from `None` to `Some`:

```rust
// After the main use_effect, add:
{
    let has_data = note_data;
    use_effect(move || {
        let is_loaded = has_data.read().is_some();
        if is_loaded {
            spawn(async move {
                crate::stores::ui::scroll_restore::set_scroll_y(0.0).await;
            });
        }
    });
}
```

### Fix 2: Defer scroll reset in Layout until after render (Defensive fix)

The Layout's scroll reset at `src/routes/mod.rs:1005` fires `set_scroll_y(0.0)` immediately on route change. For routes that render `AddressViewer` (which shows a spinner then swaps to real content), this is too early. Add a small delay or use `requestAnimationFrame` to ensure the scroll happens after the first real content render:

```rust
// src/routes/mod.rs, inside the use_effect spawn block
} else {
    // Delay scroll-to-top to allow async content viewers to render
    crate::platform::timer::sleep_ms(50).await;
    crate::stores::ui::scroll_restore::set_scroll_y(0.0).await;
}
```

### Fix 3: Scroll to top after parents render (Belt-and-suspenders)

When `parent_events` is populated (line 1029-1033), the page grows upward as parents are prepended. Add a scroll reset after parents are set:

```rust
// After parent_events.set(sorted) or parent_events.set(merged)
spawn(async move {
    crate::stores::ui::scroll_restore::set_scroll_y(0.0).await;
});
```

## Files Involved

| File | Role |
|------|------|
| `src/routes/mod.rs:987-1011` | Layout scroll restoration `use_effect` — fires too early |
| `src/routes/address_viewer.rs:11-256` | AddressViewer — shows spinner, then swaps to real content asynchronously |
| `src/components/viewers/note_viewer.rs:870-1413` | NoteViewer — loads note/parents/replies async, no scroll reset after load |
| `src/components/note_card.rs:317-333` | NoteCard click handler — `nav.push(Route::AddressViewer)` |
| `src/stores/ui/scroll_restore.rs` | Scroll restore utilities (`set_scroll_y`, `get_tracked_scroll_y`) |
| `dioxus-web-0.7.5/src/history.rs:136-151` | Dioxus WebHistory — calls `scrollTo(0,0)` on push, but before content renders |

## Environment

- Web (WASM) target
- Dioxus 0.7.5
- Affects all thread/note detail navigations from the feed

### LINKED ISSUE #350: Feature: DeFlock ALPR Camera Map (/deflock page)
### Goal

Build a `/deflock` page that displays ALPR (Automated License Plate Reader) camera locations on an interactive OSM map, similar to [maps.deflock.org](https://maps.deflock.org). Users can see camera locations, filter by operator/zone, and get directions.

### Background — How DeFlock Works

DeFlock maps ALPR cameras sourced entirely from OpenStreetMap. The OSM tagging convention is:

```
node["man_made"="surveillance"]["surveillance:type"="ALPR"]
```

Each camera node has tags like:
- `operator` — who runs it (e.g., "Flock Safety", "Local PD")
- `brand` / `manufacturer` — hardware brand
- `surveillance:zone` — "traffic", "town", "parking"
- `camera:mount` — "pole", "wall", "street_light"
- `direction` or `camera:direction` — facing direction in degrees
- `start_date` — when installed
- `ref` — reference ID
- `wikimedia_commons` — photo link

DeFlock runs an AWS Lambda that queries Overpass for all ALPR nodes worldwide, segments into tiles, and caches on CDN. Their frontend (`deflockhopper_maps`) fetches a full ~62K-camera GeoJSON dump or viewport tiles.

**Our approach:** Query Overpass directly using viewport-based bounding box queries. Self-sufficient (no dependency on DeFlock's infrastructure), always fresh, and matches nostr.blue's existing viewport-based fetch pattern from the Places module.

### Overpass API Details

- **Endpoint:** `https://overpass-api.de/api/interpreter` (primary), `https://overpass.kumi.systems/api/interpreter` (mirror/fallback)
- **Method:** GET with `?data=` query string containing the query
- **Rate limits:** 2 concurrent requests, ~10000 queries/day per IP. 180s timeout for large queries.
- **Viewport query template:**
```
[out:json][timeout:25];
node["man_made"="surveillance"]["surveillance:type"="ALPR"]({south},{west},{north},{east});
out body;
```

---

### 1. Separate module, not merged into Places

The `places/` module is about Nostr-native geotagged places (NIP-28 channels with kind 37515). ALPR cameras are OSM data with different semantics, different markers, and no Nostr integration. Mixing them would pollute the Places data model. A separate `/deflock` route keeps concerns clean.

### 2. Reuse Leaflet infrastructure, not the PlacesMapContainer component

`PlacesMapContainer` (1564 lines at `src/components/places/map_container.rs`) is tightly coupled to Nostr subscriptions, geohash fetching, BTCMap overlay, and place-specific marker/popup rendering. Copying and modifying it would create a maintenance nightmare. Instead, create a `DeflockMapContainer` that:
- Reuses the Leaflet loader JS (extract to shared utility)
- Reuses the popup CSS (`places-popup` class — already global)
- Reuses the viewport tracking pattern
- Has its own marker rendering, popup content, and data fetching

### 3. Viewport-based Overpass queries with caching

Store fetched cameras in a `GlobalSignal<HashMap<u64, AlprCamera>>` keyed by OSM node ID. On viewport change (`moveend` event), query Overpass for the new bounding box. Track which geohash regions have been fetched to avoid duplicate queries (same pattern as Places module's `FETCHED_GEOHASHES`).

### 4. No new Cargo dependencies

Everything needed is already in `Cargo.toml`:
- `reqwest` with `json` + `rustls-tls` — HTTP client for Overpass API
- `serde` / `serde_json` — deserialization
- `geohash` — viewport region tracking
- Leaflet is loaded client-side via JS eval, not a Rust crate

---

## File Structure

```
src/
├── routes/
│   ├── deflock/
│   │   ├── mod.rs              — Route module declarations
│   │   ├── home.rs             — /deflock landing page (stats, search, link to map)
│   │   └── map.rs              — /deflock/map full-screen map page
│   └── mod.rs                  — Add DeflockHome, DeflockMap to Route enum
├── components/
│   ├── deflock/
│   │   ├── mod.rs              — Module declarations
│   │   ├── map_container.rs    — Leaflet map init, viewport tracking, marker rendering
│   │   ├── camera_card.rs      — Camera detail card for list view
│   │   └── filter_bar.rs       — Operator/brand/zone filter controls
│   └── mod.rs                  — Add `pub mod deflock;`
├── services/
│   ├── deflock.rs              — Overpass API client, camera parsing, GeoJSON fallback
│   └── mod.rs                  — Add `pub mod deflock;`
├── stores/
│   ├── deflock_store.rs        — GlobalSignals for cameras, viewport, filters
│   └── mod.rs                  — Add `pub mod deflock_store;`
└── utils/
    ├── leaflet_shared.rs       — Extracted Leaflet loader JS + shared utilities
    └── mod.rs                  — Add `pub mod leaflet_shared;`
```

---

## Existing Code to Reference

These are the patterns to follow. **Read them before implementing.**

### PlacesMapContainer — `src/components/places/map_container.rs`
- **Lines 104-129:** `LEAFLET_LOAD_JS` constant — dynamically loads Leaflet CSS+JS from unpkg. Extract this to shared utility.
- **Lines 287-486:** Map initialization JS eval — sets up tiles (CARTO dark), viewport tracking (`moveend` → `window.__placesViewport`), click handler, directions (OSRM), popup CSS injection.
- **Lines 488-743:** Viewport polling loop — reads `window.__placesViewport` every 500ms, computes geohash prefixes, fetches from Nostr relays. The Deflock version replaces Nostr fetches with Overpass API calls.
- **Lines 131-228:** `build_markers_js()` — constructs JS that renders markers with SVG divIcons. Use as template for camera markers.
- **Lines 775-840:** Route polling — reads `window.__placesRouteInfo` for directions display.

### Places service — `src/services/places.rs`
- `haversine_km()` (line 160) — distance calculation, reuse directly.
- `geohash_prefix()` (line 124) and `geohash_precisions_for_zoom()` (line 132) — viewport region tracking, reuse directly.
- `fetch_osm_enrichment()` (line 262) — pattern for OSM API calls, shows how to use `http_client()` and cache results.
- `Place` struct (line 178) and `BtcMapPlace` struct (line 201) — data model patterns.

### Places store — `src/stores/places_store.rs`
- GlobalSignals pattern: `PLACES`, `VIEWPORT`, `FETCHED_GEOHASHES`, `USER_LOCATION`, `DIRECTIONS`.
- `merge_place()` — dedup pattern (key + timestamp comparison).
- `viewport_needs_refetch()` — prevents redundant fetches.

### Route registration — `src/routes/mod.rs`
- Lines 609-612: Places route enum entries.
- Lines 894-895: Places parent_route entries.
- Line 163: `use places::{PlacesHome, PlacesMap};` import pattern.

### Shared HTTP client — `src/platform/http.rs`
- `http_client()` returns `Result<&'static reqwest::Client, &'static reqwest::Error>`.
- OnceLock singleton, 15s timeout on native, no timeout on WASM (browser fetch handles it).
- Use this for all Overpass API calls.

### Geolocation — `src/platform/geolocation.rs`
- `get_current_position()` — returns `Result<(f64, f64), String>` (lat, lon).
- Works on both WASM and Android via `navigator.geolocation` JS API.

---

### AlprCamera (in `src/services/deflock.rs`)

```rust
pub struct AlprCamera {
    pub osm_id: u64,
    pub lat: f64,
    pub lon: f64,
    pub operator: Option<String>,
    pub brand: Option<String>,
    pub direction: Option<f64>,           // degrees 0-360
    pub direction_cardinal: Option<String>, // "N", "NE", etc.
    pub surveillance_zone: Option<String>, // "traffic", "town", "parking"
    pub mount_type: Option<String>,        // "pole", "wall", "street_light"
    pub ref_id: Option<String>,
    pub start_date: Option<String>,
    pub osm_timestamp: Option<String>,
    pub osm_version: Option<u32>,
    pub wikimedia_commons: Option<String>,
}
```

### BoundingBox (in `src/services/deflock.rs`)

```rust
pub struct BoundingBox {
    pub south: f64, pub west: f64, pub north: f64, pub east: f64,
}
```

With `from_center_radius(lat, lon, radius_km)` and `overlaps(&other)` methods.

---

### Task 1: Overpass API service (`src/services/deflock.rs`)

Create the data layer:
- `AlprCamera` struct (Serialize, Deserialize, PartialEq, Clone, Debug)
- `BoundingBox` struct with `from_center_radius()` and `overlaps()`
- Overpass response types (`OverpassResponse`, `OverpassElement`)
- `parse_direction()` — converts OSM direction tags (numeric or cardinal) to `(Option<f64>, Option<String>)`
- `parse_element()` — extracts camera fields from Overpass JSON element
- `fetch_cameras_in_bbox(bbox)` — queries Overpass, tries primary then fallback endpoint
- `fetch_camera_count()` — fetches worldwide count from `https://cdn.deflock.me/alpr-counts.json`
- Unit tests for bbox math, direction parsing, element parsing

Overpass query format (GET with `?data=` param):
```
[out:json][timeout:25];
node["man_made"="surveillance"]["surveillance:type"="ALPR"]({south},{west},{north},{east});
out body;
```

Register in `src/services/mod.rs` with `pub mod deflock;`.

### Task 2: Camera store (`src/stores/deflock_store.rs`)

GlobalSignals:
- `CAMERAS: GlobalSignal<HashMap<u64, AlprCamera>>` — keyed by OSM node ID for dedup
- `VIEWPORT: GlobalSignal<Option<(f64, f64, f64, f64)>>` — (lat, lon, radius_km, zoom)
- `FETCHED_GEOHASHES: GlobalSignal<HashSet<String>>` — prevents duplicate Overpass queries
- `FILTERS: GlobalSignal<CameraFilters>` — active filter settings
- `CAMERAS_LOADING: GlobalSignal<bool>`
- `LAST_ERROR: GlobalSignal<Option<String>>`
- `WORLDWIDE_COUNT: GlobalSignal<Option<u64>>`

`CameraFilters` struct with `operators`, `brands`, `zones` (HashSets), `search_query` (String), and a `matches(&AlprCamera) -> bool` method.

Helper functions: `merge_camera()`, `merge_cameras()`, `mark_geohash_fetched()`, `is_geohash_fetched()`, `get_unique_operators()`, `get_unique_brands()`, `get_filtered_cameras()`.

Register in `src/stores/mod.rs`.

### Task 3: Extract shared Leaflet utility (`src/utils/leaflet_shared.rs`)

Extract from `PlacesMapContainer`:
- `LEAFLET_LOAD_JS` constant (lines 104-129 of map_container.rs) — the promise-based Leaflet loader
- `POPUP_STYLE_JS` constant — the popup CSS injection JS (currently inline at line 321)

Update `PlacesMapContainer` to use the shared constants instead of its local copies.

Register in `src/utils/mod.rs`.

### Task 4: DeflockMapContainer (`src/components/deflock/map_container.rs`)

The core component. Pattern follows `PlacesMapContainer` but with:
- **Map init:** center `[39.5, -98.35]` (US center), zoom 4, `minZoom: 3`
- **Camera marker SVG:** Red pin (`#ef4444`) with dark camera lens circle. When direction data exists, overlay a 70° FOV cone (semi-transparent red) rotated by the direction degrees.
- **Popup content:** Operator name, zone badge (color-coded: traffic=amber, town=blue, parking=purple), direction badge, brand/mount/installed date details, OSM link button, Directions button.
- **Viewport tracking:** `moveend` → `window.__deflockViewport` (parallel to Places' `window.__placesViewport`)
- **Viewport poll loop (500ms):** Read viewport, compute geohash prefixes via `places::geohash_precisions_for_zoom(zoom)`, check `deflock_store::is_geohash_fetched()`, if not fetched → compute BoundingBox → `deflock::fetch_cameras_in_bbox()` → `merge_cameras()` → `mark_geohash_fetched()`. On error, unmark geohash for retry.
- **Marker re-render:** `use_effect` that watches `CAMERAS` signal, hashes the filtered camera set, re-renders markers via JS eval when changed.
- **Geolocation:** Reuse `crate::platform::geolocation::get_current_position()`, fly to user location on map.
- **Directions:** Reuse `window.__placesRequestDirections` JS function (OSRM routing). Add route polling loop (same pattern as PlacesMapContainer lines 775-840).
- **Cleanup:** `use_drop` removes Leaflet map instance and camera layer.

Key JS functions to set up in map init eval:
- `window.__deflockViewport` tracking
- `window.__placesEscapeHtml` (if not already defined by Places)
- `window.__placesRequestDirections` (if not already defined)

### Task 5: DeflockHome landing page (`src/routes/deflock/home.rs`)

Landing page with:
- Header with "📷 DeFlock" title and "Powered by OpenStreetMap" link
- Hero section: title, description, worldwide camera count (fetched from DeFlock CDN), locally loaded count
- "Open Camera Map" CTA button → `Route::DeflockMap {}`
- Info cards: "What are ALPR cameras?", "Data Source" (OSM tag explanation), "Report a Camera" (link to deflock.org/report)
- Footer: GitHub link

### Task 6: DeflockMap route page (`src/routes/deflock/map.rs`)

Simple wrapper — gates on `CLIENT_INITIALIZED` for UI consistency, renders `DeflockMapContainer`. Note: the Deflock map doesn't actually need Nostr relays — it queries Overpass directly. The gate is purely for consistent loading screen UX.

### Task 7: Register routes (`src/routes/mod.rs`)

- Add `pub mod deflock;` after `pub mod places;`
- Add `use deflock::{DeflockHome, DeflockMap};` after the Places use statement
- Add Route enum variants:
  ```rust
  #[route("/deflock")]
  DeflockHome {},
  #[route("/deflock/map")]
  DeflockMap {},
  ```
- Add parent_route entries:
  ```rust
  Route::DeflockHome {} => Some(Route::Explore {}),
  Route::DeflockMap {} => Some(Route::DeflockHome {}),
  ```

### Task 8: Sidebar entry (`src/stores/ui/sidebar_store.rs`)

Add Deflock to the default sidebar items. Match the existing `SidebarItem` struct pattern used for Places. Use emoji icon "📷" or a camera SVG. Route: `/deflock`.

### Task 9: Filter bar (`src/components/deflock/filter_bar.rs`)

Floating control bar on the map with:
- Search input (filters by operator/brand/zone/ref text)
- Zone toggle buttons (Traffic, Town, Parking, Other) — color-coded
- Operator dropdown (checkbox list of unique operators from loaded cameras)

Updates `deflock_store::FILTERS` on change. The map container re-renders markers when filters change via the `get_filtered_cameras()` call in the marker render effect.

### Task 10: Directions panel

Add route polling to `DeflockMapContainer` (same pattern as PlacesMapContainer lines 775-840). Display a floating panel when directions are active: distance, duration, destination name, dismiss button.

---

## Verification Steps

After all tasks, run the full pre-commit check suite from `AGENTS.md`:

```bash
# WASM (default target)
cargo check --target wasm32-unknown-unknown
cargo clippy --target wasm32-unknown-unknown -- -D warnings

# Desktop
cargo check --no-default-features --features desktop
cargo clippy --no-default-features --features desktop -- -D warnings

# Mobile
cargo check --no-default-features --features mobile
cargo clippy --no-default-features --features mobile -- -D warnings

# Tests
cargo test

# Dioxus check
dx check
```

---

### Overpass rate limits
Overpass allows ~2 concurrent requests and ~10000/day per IP. Geohash dedup prevents re-fetching the same area. The 500ms poll interval debounce prevents rapid-fire. Fallback endpoint provides resilience. Consider a minimum zoom threshold (don't query below zoom 8 — viewport too large for Overpass timeout).

### WASM HTTP timeout
`http_client()` has no timeout on WASM. Overpass `[timeout:25]` helps server-side. The geohash is marked fetched *before* the async fetch, preventing duplicate queries. On failure, the geohash is unmarked for retry.

### No marker clustering
Dense areas (city centers) could have hundreds of markers. Leaflet degrades past ~500 visible. Future: add `Leaflet.markercluster`. For now, filters help narrow display, and viewport-based fetch limits density.

### `std::time::Duration` on WASM
The viewport poll loop uses `std::time::Duration`. This is fine because it's passed to `crate::platform::timer::sleep()` which uses `web_time` internally, not `std::thread::sleep()`. Do NOT use `std::thread::sleep` or `tokio::time::sleep` on WASM.

### Dioxus hooks must be called unconditionally
All `use_signal`, `use_effect`, `use_drop` calls must be at the top of the component function body — never behind conditional early returns.

---

## Future Enhancements (Out of Scope)

1. **Privacy routing** — Calculate routes that avoid camera clusters (requires custom OSRM profiles or Valhalla routing engine)
2. **Timeline mode** — Show camera installation dates with a time slider
3. **Density heatmap** — Leaflet heatmap layer for camera density
4. **Community reporting** — Users publish Nostr notes about specific cameras (geotagged)
5. **Marker clustering** — `Leaflet.markercluster` for dense areas
6. **Offline caching** — Store fetched camera data in IndexedDB

---

A full implementation plan with complete code for all tasks is saved at `docs/plans/2026-07-21-deflock-alpr-map.md` in the repo.

### LINKED ISSUE #351: fix(relay): indexer relays never added to pool at startup — profile metadata fails to load on home feed
## Bug Description

Profile metadata (kind 0) fails to load for followed accounts on the home feed. Users see truncated hex pubkeys instead of display names, avatars, and other profile data. The feed events themselves load correctly — only the author metadata is missing.

This is a reintroduction of the bug fixed in `2a8559b4` (fix(metadata): fix profile loading for follows, feed authors, and mentions). The profile-fetching machinery is all correct — the indexer relays that serve kind 0 events are never added to the connection pool at startup.

## Root Cause

There are two functions that read the indexer relay list, and they behave differently when the `INDEXER_RELAYS` signal is empty (which it is at startup):

**`add_indexer_relays_to_client`** (`src/stores/relay/nip65.rs:891`) reads the **raw signal directly**:

```rust
pub async fn add_indexer_relays_to_client(client: Arc<Client>) {
    let indexer_urls = INDEXER_RELAYS.peek().clone();  // ← empty at startup
    if indexer_urls.is_empty() {
        return;  // ← silently no-ops, adds nothing to the pool
    }
    // ...
}
```

**`fetch_events_from_indexers`** (`src/stores/relay/nip65.rs:927`) uses `get_indexer_relay_urls()`, which **has the correct fallback**:

```rust
pub fn get_indexer_relay_urls() -> Vec<String> {
    let relays = INDEXER_RELAYS.peek().clone();
    if relays.is_empty() {
        default_indexer_relays()  // ← returns the 6 hardcoded defaults
    } else {
        relays
    }
}
```

The `INDEXER_RELAYS` signal is initialized as `Vec::new()` and is only populated by `init_private_relay_lists` (which parses a gift-wrapped kind 10086 event). That function runs **late** in `run_post_login_init` — long after `add_indexer_relays_to_client` was supposed to add the indexers to the pool.

## The Broken Startup Sequence

In `run_post_login_init` (`src/stores/auth_store.rs:894-960`):

1. `INDEXER_RELAYS` = `Vec::new()` (empty)
2. Line 904: `add_indexer_relays_to_client(client)` → reads empty signal → **returns immediately, adds zero relays to the pool**
3. Line 910: `client.connect()` → connects general/user relays only, **no indexers in the pool to connect**
4. Line 949: `warmup_profiles_from_network` spawned → calls `wait_for_indexer_connected` → polls the pool for 15s looking for indexers that were **never added** → times out
5. `fetch_profiles_batch_native` → calls `fetch_events_from_indexers` → `get_indexer_relay_urls()` correctly returns the 6 default URLs → checks `pool().all_relays()` for those URLs → **none exist in pool** → returns `Err("No indexer relays connected yet")`
6. All follow metadata marked as errored (retryable, not exhausted), but retries hit the same wall
7. NoteCard's `queue_profile_request` → `drain_profile_queue` → same `fetch_events_from_indexers` → same `Err` → **profiles never load**

The profile exhaustion cooldown (5 min, 2 attempts) eventually kicks in and suppresses retries, making the missing metadata persistent for the session.

## Impact

- Every cold login shows hex keys instead of names/avatars for any followed account not already in the SDK local database
- The DB-only warmup (`warmup_profiles_from_db`) masks this for returning users with cached profiles, but new follows or fresh installs see the full breakage
- The `wait_for_indexer_connected` call in `warmup_profiles_from_network` burns 15 seconds on every login waiting for indexers that can never connect

## Suggested Fix

Make `add_indexer_relays_to_client` use `get_indexer_relay_urls()` instead of reading the raw signal, so it falls back to the default indexer list when the user hasn't published a custom kind 10086:

```rust
pub async fn add_indexer_relays_to_client(client: Arc<Client>) {
    let indexer_urls = get_indexer_relay_urls();  // ← was: INDEXER_RELAYS.peek().clone()
    if indexer_urls.is_empty() {
        return;
    }
    log::info!("Adding {} indexer relays as DISCOVERY-only", indexer_urls.len());
    for url_str in &indexer_urls {
        if let Ok(url) = RelayUrl::parse(url_str) {
            let _ = client.add_discovery_relay(url).await;
        }
    }
}
```

This is a one-line change. It aligns `add_indexer_relays_to_client` with the already-correct `fetch_events_from_indexers` and `fetch_own_lists_from_indexers` (both of which already use `get_indexer_relay_urls()` or the same fallback pattern).

When `init_private_relay_lists` later populates `INDEXER_RELAYS` with the user's custom list, those relays should also be added to the pool at that point (or the pool should be reconciled). But the defaults must be present from the first `connect()` call so metadata fetches work immediately.

## Verification

After the fix, a cold login should show:
- Log: `Adding 6 indexer relays as DISCOVERY-only` (not zero)
- `wait_for_indexer_connected` succeeds within seconds instead of timing out at 15s
- `warmup_profiles_from_network: loaded N profiles` with N > 0
- Home feed shows display names and avatars instead of hex keys

patrickulrich:
Bundles the new DeFlock ALPR camera map feature with two standalone bug fixes.
Closes #348, closes #350, closes #351.

### ✨ New: DeFlock ALPR camera map (`/deflock`)
Adds an interactive Leaflet map of Automatic License Plate Reader (ALPR) camera
locations from OpenStreetMap (Overpass API), modeled on maps.deflock.org.

- Full-screen map at `/deflock` (route collapsed; no separate landing page)
- Zoom-gated direction cones (simple circles at z<14; DeFlock-accurate FOV-cone
  SVGs at z≥14, with range/multi-value direction parsing)
- **Leaflet.markercluster** for thousands of markers without jank
- **IndexedDB persistence** (`deflock_cache_db.rs`) — revisiting an area next
  session is instant
- **Bbox coverage tracking** — skips refetch when the viewport is a subset of a
  previously-fetched bbox
- **Parallel mirror racing** across 3 Overpass endpoints (first response wins)
- Filter bar (zone/operator/text) + floating OSRM directions panel
- New `SidebarItem::Deflock` (visible by default, EyeIcon); Leaflet loader
  shared via `utils/leaflet_shared.rs` with Places

### 🐛 Fix: thread opens scrolled down instead of at top (#348)
Threads opened mid-scroll because the Layout's scroll-to-zero fired against the
AddressViewer spinner before NoteViewer's async content (parents + note +
replies) painted, and browser scroll-anchoring then shifted the viewport.

- New `PENDING_SCROLL_TARGET` global: Layout sets `0.0` (forward push) or
  `saved-y` (Back/popstate); NoteViewer applies it in a post-paint `use_effect`,
  generation-guarded against stale fires.
- `overflow-anchor` disabled on `<html>` during load (re-keyed per note),
  restored when `loading_replies` settles; `spawn_forever` restore on unmount.
- **Result:** feed/reply-in-thread → opens at top; Back → restores exact prior
  position; streaming replies no longer yank to top.

### 🐛 Fix: relay lists never seeded at boot, eliminating USER_RELAYS_APPLIED race (#351)
Indexers were missing from the pool at startup (`add_indexer_relays_to_client`
read an empty `INDEXER_RELAYS` before late init), so kind-0 metadata silently
failed on every cold login. Now seeds all public relay lists from disk
(localStorage mirror + SDK DB) before the first `connect()`, so
`USER_RELAYS_APPLIED` flips synchronously — turning 44 `wait_for_user_relays`
call sites + 11 gate readers into instant no-ops.


src/utils/leaflet_shared.rs
```diff
@@ -0,0 +1,78 @@
...
+});
+"#;
+
+/// CSS that styles Leaflet popups with the nostr.blue dark theme.
+/// Injects once per page via the `window.__placesPopupStyleAdded` flag.
+/// Note: brace-doubled for direct interpolation into Rust `format!()` strings.
+pub const POPUP_STYLE_JS: &str = r#"
+if (!window.__placesPopupStyleAdded) {{
+    const style = document.createElement('style');
+    style.textContent = '.places-popup .leaflet-popup-content-wrapper{{background:rgba(20,20,20,0.92);backdrop-filter:blur(10px);border-radius:12px;border:1px solid rgba(255,255,255,0.1);box-shadow:0 8px 32px rgba(0,0,0,0.5);color:#e5e5e5;padding:0;}} .places-popup .leaflet-popup-content{{margin:12px 16px;line-height:1.4;}} .places-popup .leaflet-popup-tip{{background:rgba(20,20,20,0.92);}} .places-popup .leaflet-popup-close-button{{color:#737373 !important;font-size:18px !important;top:6px !important;right:8px !important;}} .places-popup .leaflet-popup-close-button:hover{{color:#fff !important;}}';
```
goodmorningbitcoin: **High — literal `{{` in generated CSS breaks popup styling (regression)**

`POPUP_STYLE_JS` keeps its braces doubled (`{{` / `}}`), which was correct when this text was *inlined* in a `format!()` template literal (the old Places code). It is now substituted as a named argument (`popup_style = POPUP_STYLE_JS` in both map containers), and `format!()` only unescapes braces in the *template* — never in substituted *values* (verified with a standalone rustc repro). The evaluated JS therefore contains literal `{{...}}` inside the injected `textContent` CSS; the browser drops those rules, so the dark-theme popup styling silently regresses for **both** Places and Deflock.

Fix: use single braces in the constant (it no longer needs escaping), and note in the doc comment that it must not be re-inlined into a brace-escaped template.

src/components/deflock/map_container.rs
```diff
@@ -0,0 +1,638 @@
...
+                    }};
+                }});
+
+                if (!window.__placesEscapeHtml) {{
+                    window.__placesEscapeHtml = function(str) {{
+                        if (!str) return '';
+                        return str.replace(/&/g,'&amp;').replace(/</g,'&lt;').replace(/>/g,'&gt;').replace(/"/g,'&quot;').replace(/'/g,'&#39;');
+                    }};
+                }}
+                if (!window.__placesRequestDirections) {{
```
goodmorningbitcoin: **High — `__placesRequestDirections` can capture the wrong map instance**

This global is defined by both Places (`places/map_container.rs:352`, assigned unconditionally on map init) and here (guarded `if (!window.__placesRequestDirections)`). Whichever map initializes first permanently owns the handler, and the closure captures *that* map's `map` variable. Visiting Places before Deflock means the Directions button on Deflock popups draws the polyline + `fitBounds` against the (removed) Places map — nothing renders on the Deflock map. `__placesRouteLayer` is likewise shared, so clearing a route on one map mutates the other's state.

Suggested fix: dispatch by active container id (e.g. `window.__placesRequestDirectionsFor(mapId, ...)` reading the map from `window.leafletMaps`), or namespace a separate `__deflockRequestDirections`.

src/services/deflock.rs
```diff
@@ -0,0 +1,354 @@
...
+}
+
+impl BoundingBox {
+    pub fn from_center_radius(lat: f64, lon: f64, radius_km: f64) -> Self {
+        let lat_delta = radius_km / 111.0;
+        let lon_delta = radius_km / (111.0 * lat.to_radians().cos().max(0.01));
+        Self {
+            south: (lat - lat_delta).max(-90.0),
+            north: (lat + lat_delta).min(90.0),
+            west: (lon - lon_delta).max(-180.0),
```
goodmorningbitcoin: **Medium — antimeridian clamp produces an inverted bbox**

`from_center_radius` clamps west to >= -180 and east to <= 180 independently. Near the antimeridian (e.g. lon 179.9 with a +-1deg radius) the west edge wraps below -180 and clamps to -180 while east clamps at 180 — yielding west=-180 / east=180, i.e. a whole-world Overpass query. `is_viewport_covered`'s containment check (assumes west < east) can then never mark it covered, so NZ/Fiji/Alaska-edge viewports refetch a max-size query on every pan.

Suggested fix: after computing the bounds, if `west > east` (wrap detected), either skip the fetch or split into two bboxes west of/east of the line.

src/components/deflock/filter_bar.rs
```diff
@@ -0,0 +1,93 @@
...
+    let search = current_filters.search_query.clone();
+
+    rsx! {
+        div { class: "fixed bottom-20 left-4 right-4 z-[60] bg-black/70 backdrop-blur-md rounded-xl p-3 max-w-md mx-auto pointer-events-auto",
+            input {
+                class: "w-full bg-white/5 border border-white/10 rounded-lg px-3 py-2 text-sm text-white placeholder:text-white/40 mb-2",
+                r#type: "text",
+                placeholder: "Search cameras...",
+                value: "{search}",
+                oninput: move |e| {
```
goodmorningbitcoin, line 27: **Low — no debounce on search input**

Every keystroke writes `FILTERS`, and the marker effect rebuilds the entire `markerClusterGroup` (remove layer -> re-add all filtered markers) per change. Fine at a few hundred cameras; with a warmed 50k-camera IndexedDB cache this becomes per-keystroke jank. The codebase already has debounce patterns (the profile queue batches at 200ms) — worth ~250ms here.

src/components/ppq_import_form.rs
```diff
@@ -0,0 +1,139 @@
...
+        }
+    }
+}
+
+async fn import_account(
+    credit_id: &str,
+    pasted_key: Option<String>,
+) -> Result<PpqAccountState, String> {
+    if let Some(api_key) = pasted_key {
+        let credit_id = ppq::validate_credit_id(credit_id).await?;
```
goodmorningbitcoin: **Low — pasted API key is stored unvalidated**

The pasted-key path validates only the Credit ID (`GET /keys` with `x-credit-id`); the key itself is never checked against PPQ. A typo'd key becomes `managed_api_key` and the user only discovers it at the first chat call with a confusing 401. Either validate the key with a cheap authorized request on import, or set the expectation in the UI copy ("key will be verified on first use").

src/stores/relay/persistence.rs
```diff
@@ -0,0 +1,335 @@
...
+    let events = client.database().query(filter).await.ok()?;
+    let mut relays = Vec::new();
+    let mut dm_relays = Vec::new();
+    let mut updated_at = 0u64;
+    for event in events.into_iter() {
+        match event.kind.as_u16() {
+            10002 => {
+                let parsed = parse_relay_list_event(&event);
+                if !parsed.is_empty() {
+                    relays = parsed;
```
goodmorningbitcoin: **Medium — seed fold is last-iteration-wins, not newest-by-created_at**

`collect_metadata_from_db` folds the (<=10) queried events with plain assignment — the *last iterated* kind 10002 wins, not the newest. `database().query()` ordering is not a documented newest-first guarantee across backends (IndexedDB vs NDB), so an older relay-list snapshot can win the seed. The profile code added in this same PR does this correctly with `max_by_key(|e| e.created_at)` — mirror that here (per kind). Also removes the shared `limit(10)` concern across kinds 10002+10050.

goodmorningbitcoin (commented):
## 🤖 CodeRabbit-Style Review — Part 1

**Reviewed at:** `bd6fa787` (includes all 9 commits: Deflock feature, relay-seed + scroll fixes, minicards, pinboard NIP compliance, PPQ import, profile metadata, PWA/SW)

### Walkthrough

This release PR bundles five mostly-independent changes:

1. **DeFlock ALPR map** (`/deflock`) — new Overpass-backed Leaflet map with markercluster, IndexedDB persistence, bbox-coverage dedup, and 3-endpoint mirror racing.
2. **Relay-list seeding** (#351) — two-tier (localStorage mirror + SDK DB) seed before first `connect()` so `USER_RELAYS_APPLIED` flips synchronously; also fixes `add_indexer_relays_to_client` reading an empty signal.
3. **Thread scroll** (#348) — `PENDING_SCROLL_TARGET` applied post-paint in NoteViewer, `overflow-anchor` suppression during load.
4. **Pinboards** — NIP-01/33 compliance: derived `d` tags on kind-39067 pins, A-tag merge on re-pin, Unicode slugify, media kind inference.
5. **PPQ Credit-ID import** — UUID-validated import with key reuse/creation; plus the new profile-metadata freshness work and the SW network-first fix.

### Findings — high level

| Severity | Count | Summary |
|---|---|---|
| 🟠 High | 2 | `format!` brace regression silently breaks Places+Deflock popup CSS; shared `__placesRequestDirections` global can capture the wrong map |
| 🟡 Medium | 3 | Scroll-to-top refire on same-note refresh; relay-seed fold is last-iteration-wins; antimeridian bbox inversion; defaults can durably overwrite a user's relay list (4 total, one in Part 2 batch) |
| 🔵 Low | 5 | Un-`max_by_key`'d metadata `next()` calls; search-input debounce; unvalidated pasted PPQ key; `created_at: None` cache-regression window; fold tie non-determinism |
| ⚪ Nitpick | 2+ | Kind-map drift; IPv6 loopback dev detection; see also Part 2 |

Inline comments below (Part 1 = new files; continued in Part 2 for modified files).

### Additional observations (not inline — lines outside diff hunks or cross-cutting)

- **`profile_to_metadata` drops `lud06`** (and `bot`/`birthday`) on the tier-1 cache → viewer round-trip. `lud16` survives, so zaps work for modern profiles, but legacy `lud06`-only profiles lose their zap fallback when served from `PROFILE_CACHE`.
- **Outbox rescue ordering** — `rescue_candidates` takes the first `MAX_OUTBOX_RESCUE` of `still_missing` in batch order; if those keep failing, later pubkeys starve across retries. A rotate/shuffle would spread the budget.
- **`controllerchange` auto-reload** (sw-register.js) force-reloads the app mid-session when the hourly update check finds a new SW — in-progress compose state is lost. Pre-existing behavior, but now that updates actually propagate (network-first), users will hit it more often.
- **Overpass etiquette** — racing all 3 mirrors in parallel on *every* uncovered viewport triples request volume per fetch. `select_all` + drain-on-error is implemented correctly; consider sequential-with-timeout as a gentler default.

### ✅ Looks good

- Relay seeding architecture is sound: collect-then-write split avoids the `WritableRef`-across-`await` panic, gift-wrap exclusion avoids boot-time NIP-07 prompts, and the early `USER_RELAYS_APPLIED` flip genuinely kills the race class.
- Pinboard `d`-tag derivation matches the addressable-event semantics (derived from referenced content, coordinate normalization incl. naddr), and the A-tag union on re-pin fixes silent board-eviction. Tests cover the legacy paths.
- PPQ import: UUID validation is strict (no header injection surface), response bodies stay redacted to byte counts, and key-name collision handling (including revoked-key name squatting) is thorough.
- Profile freshness work: `newest_metadata_by_author` fold, blank-content-as-wipe semantics, `displayName` camelCase fallback, and the bounded/concurrent outbox rescue are all well-reasoned and well-tested.
- SW: cache bump + network-first for non-hashed wasm artifacts + dev-registration guard is the right shape.

*Note: local `cargo check` of the PR branch was resource-killed on the review machine; findings are read-verified against the diff (CI is green on the PR).*

---

*Review effort: medium-high · 2 high · 4 medium · 5 low · 3 nitpicks*

> From Overpass the cameras came,
> relays seeded, threads reset —
> but braces doubled in a string
> unstyle what they once set.

goodmorningbitcoin (commented):
Re-posting inline comments from review 4939882354 (dropped due to API position format).

goodmorningbitcoin (commented):
## 🤖 CodeRabbit-Style Review — Part 2 (modified files)

Inline findings for modified files continue here (Part 1 covered the new files + walkthrough table).

### ✅ Verified-fixed notes (second-pass on the two new commits)

- **Profile freshness (`103da8af`)**: `newest_metadata_by_author` fold before cache put, blank-content-as-wipe parsing, `displayName` camelCase fallback, bounded (`MAX_OUTBOX_RESCUE=10`) + width-3 outbox rescue, and `created_at`-tagged metadata throughout — all sound and well-tested. Two low-severity residuals inline.
- **PWA/SW (`bd6fa787`)**: network-first for non-hash-named wasm + dev-registration guard is the right shape. Nitpick inline re IPv6 loopback.

*Continued: see inline comments on this review for the remaining findings.*

src/components/viewers/note_viewer.rs
```diff
@@ -1216,13 +1216,93 @@ pub fn NoteViewer(
...
+            spawn(async move {
+                // Yield one frame so DB-parent writes (set a few lines after
+                // `note_data` in the main load effect) land before the reset.
+                crate::platform::timer::sleep_ms(16).await;
+                if *load_generation.peek() != gen {
+                    return;
+                }
+                let target = crate::stores::ui::scroll_restore::PENDING_SCROLL_TARGET
+                    .peek()
+                    .unwrap_or(0.0);
```
goodmorningbitcoin: **Medium — `unwrap_or(0.0)` re-scrolls to top on same-note refresh**

The C1 effect re-fires on every `None -> Some` transition of `note_data`, including the background refresh that rewrites `note_data` for the *same* note (`fetch_main_note`, ~line 994 — e.g. prefetched -> confirmed swap, or an edited-note refetch). On that re-fire `PENDING_SCROLL_TARGET` has already been consumed (None), so the `unwrap_or(0.0)` fallback scrolls to top — yanking a mid-read user back to the top of the thread. That's the same symptom this PR fixes (#348), reintroduced on the refresh path.

Suggested fix — only apply when a target is actually pending:

```rust
let Some(target) = PENDING_SCROLL_TARGET.peek() else { return; };
set_scroll_y(target).await;
*PENDING_SCROLL_TARGET.write() = None;
```

src/stores/auth_store.rs
```diff
@@ -928,15 +950,33 @@ fn run_post_login_init() {
...
+            );
+        }
+
         // Fetch the user's NIP-65 relay list from connected relays.
+        // This is now a NETWORK REFRESH — the disk seed already populated the
+        // pool and signals. It overwrites via last-write-wins reconciliation.
         if let Err(e) = crate::stores::relay::init_user_relay_lists(client.clone()).await {
             log::warn!("Failed to load user relay lists: {}", e);
         }
+        crate::stores::relay::persistence::persist_public_relay_lists();
```
goodmorningbitcoin: **Medium — defaults can durably overwrite the user's relay list**

`persist_public_relay_lists()` runs right after `init_user_relay_lists` regardless of Ok/Err. On Err (offline boot, relays unreachable), `init_user_relay_lists` writes the *default* relay set into `USER_RELAY_METADATA` (nip65.rs:526), which this persist then mirrors into the durable tier-1 localStorage — overwriting the user's real list. Because the seed prefers the mirror (`mirror_metadata.or(db_metadata)`), subsequent boots keep serving defaults even when the SDK DB still holds the real kind 10002. One bad boot permanently replaces the user's relay list with defaults.

Suggested fix: skip this persist when `init_user_relay_lists` returned Err, or gate the mirror write on `updated_at` being newer than what the mirror already holds.

src/stores/profiles.rs
```diff
@@ -415,15 +462,46 @@ fn empty_profile(pubkey: &str) -> Profile {
...
+/// iterating arrival order would let an older version overwrite a newer one.
+pub(crate) fn newest_metadata_by_author(events: Vec<Event>) -> Vec<Event> {
+    let mut newest: HashMap<nostr::PublicKey, Event> = HashMap::new();
+    for event in events {
+        match newest.entry(event.pubkey) {
+            std::collections::hash_map::Entry::Vacant(slot) => {
+                slot.insert(event);
+            }
+            std::collections::hash_map::Entry::Occupied(mut slot) => {
+                if event.created_at > slot.get().created_at {
```
goodmorningbitcoin: **Low — tie on `created_at` makes the fold order-dependent**

Replaceable events can share a `created_at` (same-second republish). With strictly-greater `>`, the first-arriving event wins ties, so the fold's result depends on iteration order — a nondeterministic pick between two different same-second snapshots. Cheap fix: tie-break on event id (`event.id > slot.get().id`) so the fold is a total, deterministic order.

src/components/viewers/profile_viewer.rs
```diff
@@ -272,12 +273,25 @@ pub fn ProfileViewer(pubkey: String) -> Element {
...
+                // revalidate). When the underlying kind-0 event is older
+                // than 24h, race indexers/outbox in the background and
+                // replace only on a strictly-newer result.
+                if let Some(cached) = profiles::get_cached_profile(&hex_pubkey) {
                     if *rid.peek() != current_id {
                         return;
                     }
                     log::debug!("Loaded profile metadata from LRU cache");
-                    profile_data.set(Some(metadata));
+                    let displayed_created_at = cached.event_created_at.unwrap_or(0);
```
goodmorningbitcoin: **Low — `event_created_at: None` disables the strictly-newer guard**

Cached profiles built via `metadata_to_profile` (pre-PR cache entries, or any `cache_profile(..., None)` caller) have `event_created_at = None` -> `displayed_created_at = 0`, so the background revalidation replaces the cache with *whatever* the race returns — even an older snapshot than the one already displayed. For these entries the revalidation is effectively unconditional. Consider using `fetched_at` as a floor stand-in so `created_at > floor` still filters stale winners.

src/stores/nostr_client/fetching.rs
```diff
@@ -883,12 +900,13 @@ pub async fn fetch_metadata_targeted(
             )
             .await;
             relay::coverage::cleanup_ephemeral_relays(&client, &ephemeral.newly_added).await;
+            // `Events` iterates newest-first, so `next()` is the newest event.
             if let Ok(events) = result {
                 if let Some(event) = events.into_iter().next() {
```
goodmorningbitcoin: **Low — `next()` rests on a newest-first ordering assumption**

The targeted path (here) and both broadcast fallbacks take `events.into_iter().next()` relying on `Events` iterating newest-first, while the indexer path in this same file defensively uses `max_by_key(|e| e.created_at)`. If that ordering assumption ever changes, these paths silently prefer an older kind 0. Using `max_by_key` in all three spots costs nothing and removes the assumption entirely.

src/stores/social/pin_boards_store/mod.rs
```diff
@@ -196,6 +223,7 @@ impl PinReference {
                         30009 => PinContentType::Badge,
                         30067 => PinContentType::Pinboard,
                         31337 | 32267 => PinContentType::Music,
+                        39701 => PinContentType::Link,
```
goodmorningbitcoin, line 226: **Nitpick — kind->content-type maps have drifted**

Three parallel maps now exist: `infer_content_type` (this one), `infer_content_type_from_kind` (pin_new.rs), and `extract_event_metadata` (mod.rs). This PR adds 39701/0 to all of them, but 20/21/22 (Image/Video) only to the latter two — a kind-20 coordinate pin still infers `Note` here while its metadata path says `Image`. Consider collapsing into one shared `kind_to_pin_content_type(kind) -> Option<PinContentType>` consumed by all three sites.

public/sw-register.js
```diff
@@ -1,38 +1,49 @@
...
-
-                // Check for updates periodically
-                setInterval(() => {
-                    reg.update().catch(err => {
-                        console.error('[PWA] Service Worker update check failed:', err);
-                    });
-                }, 60 * 60 * 1000); // Check every hour
+    // Never let the service worker control local development: its cache-first
+    // asset strategy would serve stale wasm/js across `dx serve` rebuilds.
+    const isLocalDev = ['localhost', '127.0.0.1'].includes(location.hostname);
```
goodmorningbitcoin, line 5: **Nitpick — dev-host detection misses IPv6 loopback**

`['localhost', '127.0.0.1']` doesn't cover `http://[::1]:8080` (browsers report `location.hostname` as `[::1]`), where the SW would still register and serve cached assets across `dx serve` rebuilds. Add `'[::1]'` to the list.

src/components/places/map_container.rs
```diff
@@ -195,7 +174,7 @@ fn build_markers_js(id_json: &str, markers_json: &str) -> String {
                         ${{safeWeb ? '<div style="margin-bottom:3px;"><a href="' + safeWeb + '" target="_blank" rel="noopener" style="color:#a78bfa;text-decoration:none;font-size:12px;word-break:break-all;">🌐 ' + safeWeb.replace(new RegExp("^https?://"), "") + '</a></div>' : ''}}
                         <div style="display:flex;gap:6px;margin-top:8px;">
                             ${{s(m.naddr) ? '<a href="/' + s(m.naddr) + '" style="padding:5px 14px;border-radius:6px;background:transparent;color:#a78bfa;border:1px solid #7c3aed;cursor:pointer;font-size:12px;font-weight:500;text-decoration:none;">View Details</a>' : ''}}
-                            <button onclick="window.__placesRequestDirections(${{m.lat}},${{m.lng}},'${{name.replace(/'/g, "\\\\'")}}')"
+                            <button onclick="window.__requestDirectionsFor(${{mapId}},${{m.lat}},${{m.lng}},'${{name.replace(/'/g, "\\\\'")}}','#7c3aed')"
```
goodmorningbitcoin: **High — unquoted `${mapId}` breaks the Directions button (regression from the per-map dispatch fix)**

`mapId` is a raw string like `places-map-1755…-0`. Interpolating it bare into the inline onclick yields `window.__requestDirectionsFor(places-map-1755…-0, 39.5, …)` — parsed as arithmetic on undefined identifiers, ReferenceError at click time. Quote it: `window.__requestDirectionsFor('${mapId}',…)` (id is alphanumeric + hyphens, no escaping needed). Same fix needed at line 193 and in `deflock/map_container.rs:132`.

src/components/deflock/map_container.rs
```diff
@@ -0,0 +1,627 @@
...
+                if (refId) details += `<div style="color:#a3a3a3;font-size:12px;margin-bottom:2px;">Ref: ${{refId}}</div>`;
+
+                const popup = `<div style="min-width:220px;max-width:280px;color:#e5e5e5;font-family:-apple-system,BlinkMacSystemFont,sans-serif;font-size:13px;line-height:1.4;">
+                    <div style="font-size:15px;font-weight:600;color:#fff;margin-bottom:2px;">📷 ${{operator}}</div>
+                    <div style="margin-bottom:6px;">${{badges}}</div>
+                    <div style="border-top:1px solid rgba(255,255,255,0.1);margin:6px 0;"></div>
+                    ${{details}}
+                    <div style="display:flex;gap:6px;margin-top:8px;">
+                        <a href="${{osmUrl}}" target="_blank" rel="noopener" style="padding:5px 14px;border-radius:6px;background:transparent;color:#a78bfa;border:1px solid #7c3aed;cursor:pointer;font-size:12px;font-weight:500;text-decoration:none;">OSM</a>
+                        <button onclick="window.__requestDirectionsFor(${{mapId}},${{c.lat}},${{c.lon}},'Camera ${{c.osm_id}}','#ef4444')"
```
goodmorningbitcoin: **High — same unquoted `${mapId}` regression here**

Bare `${mapId}` in the onclick → ReferenceError at click. Quote it: `'${mapId}'`. (The `__clearRouteFor({id_json})` call in the close-button handler is correct — quoted at the Rust level.)

goodmorningbitcoin (commented):
## 🔁 Second-Pass Review (fix commits `be83f4d6`…`e745bd15`)

### ✅ Verified fixed (11 of 13)

| # | Original finding | Verdict |
|---|---|---|
| 1 | 🟠 `POPUP_STYLE_JS` doubled braces → broken popup CSS | ✅ Single braces + doc warning + regression test asserting no `{{` in either constant |
| 2 | 🟠 `__placesRequestDirections` wrong-map capture | ✅ Replaced by `__requestDirectionsFor(mapId, …)` resolving the map from `window.leafletMaps` at call time; route layer moved onto the map instance — **but see new High below** |
| 3 | 🟡 Scroll-to-top refire on same-note refresh | ✅ `let Some(target) = … else { return; }` — applies only when a target is pending |
| 4 | 🟡 Defaults durably overwriting user relay list | ✅ Persist now gated on `Ok` + `should_persist_metadata` blocks `updated_at == 0` (defaults sentinel) from overwriting a real mirror; unit-tested |
| 5 | 🟡 Relay-seed fold last-iteration-wins | ✅ `newest_relay_list_events` folds newest-by-`created_at` with id tie-break; unit-tested |
| 6 | 🟡 Antimeridian bbox inversion | ✅ `bboxes_for_center_radius` normalizes lon and emits a companion bbox for the wrapped side; 4 new tests incl. both wrap directions |
| 7 | 🔵 Unvalidated pasted PPQ key | ✅ `validate_credit_id` now attaches the Bearer header and maps 401/403 to "PPQ rejected this API key" |
| 8 | 🔵 Filter-bar search debounce | ✅ 250ms debounce with local echo; toggles stay immediate |
| 9 | 🔵 `next()` newest-first ordering assumption | ✅ All three call sites now use `min()` under `Ord for Event` (verified against nostr-0.44.3: descending `created_at`, then ascending id) |
| 10 | 🔵 `event_created_at: None` guard bypass | ✅ `fetched_at` floor stand-in |
| 11 | 🔵 Fold tie non-determinism (profiles) | ✅ Tie-break on smaller id, consistent with SDK `Ord` |
| — | `lud06` drop in `profile_to_metadata` (summary note) | ✅ Also fixed |

### ❌ New issue introduced by the fix

🟠 **High — Directions buttons throw ReferenceError on both maps** (`places/map_container.rs:177`, `:193`; `deflock/map_container.rs:132`)

The popup template literals interpolate the map id **unquoted**:

`<button onclick="window.__requestDirectionsFor(${mapId},${m.lat},…)"`

`mapId` holds a raw string like `places-map-1755…-0`; interpolating it bare into the inline `onclick` attribute yields `window.__requestDirectionsFor(places-map-1755…-0, 39.5, …)`, which evaluates as arithmetic on undefined identifiers and throws at click time. Every popup Directions button on both maps is now dead.

Fix: quote it — `window.__requestDirectionsFor('${mapId}',…)` (the id is alphanumeric + hyphens, so no escaping concerns), or JSON-encode the id once at the top of the builder. The `__clearRouteFor({id_json})` call in Deflock's close button does this correctly via the quoted Rust-level substitution — same treatment needed here.

### ⏭️ Not addressed (both nitpick-level, fine to defer)

- Kind→content-type map drift (nit) — 20/21/22 still missing from `infer_content_type`'s Coordinate arm
- IPv6 `[::1]` dev-host detection (nit)

*Second-pass verdict: 11/13 verified fixed, 1 regression to fix before merge (blocking), 2 nits deferred. The regression is a one-line quoting fix on three call sites.*

goodmorningbitcoin (approved):
## 🔁 Third-Pass Review — ✅ Verified fixed (`be6fb62e`)

**Regression resolved.** All three Directions onclick sites (`places/map_container.rs:177`, `:193`, `deflock/map_container.rs:132`) now interpolate the map id as a quoted JS string (`'${mapId}'`); `git grep` confirms zero bare interpolations remain, `mapId` is still defined at the IIFE top in both builders, and two new regression tests (`test_directions_onclick_quotes_map_id`) lock it in — **both pass**.

Verified beyond the diff: the places brace-reindent hunk is token-identical to the prior version (parsed both files and diffed token streams — only divergence is the appended test module), and the full test suite runs green on this head: **945 passed, 0 failed**.

### Final tally across all three review rounds

| Round | Findings | Status |
|---|---|---|
| 1 (`bd6fa787`) | 13 findings: 2 High, 4 Medium, 5 Low, 2 Nit | 11 fixed, 2 nits deferred |
| 2 (`e745bd15`) | 1 regression (unquoted `mapId`) | ✅ fixed + regression tests |
| 3 (`be6fb62e`) | — | clean |

Nothing blocking remains. 🚀

Issue #353, opened 2026-08-09
Pinboard NIP compliance: missing d-tags on pins, board d-tag bugs, content-type inference gaps


goodmorningbitcoin:
## Problem

The pinboard implementation (kind 30067/39067) is out of compliance with the updated Pinboard NIP. The updated spec was reviewed against the codebase and four categories of problems were found — ranging from spec violations that affect every pin published to missing content-type inference.

---

## 1. CRITICAL — Kind 39067 pins have no `d` tag (spec violation)

The updated NIP lists `d` as a **required** tag for kind 39067 pins:

> **Required Tags:**
> | Tag | Description |
> |-----|-------------|
> | `d` | Unique identifier (used to generate `naddr`) |

The spec defines the `d`-tag value as the referenced content's identifier:
- `e` tag reference → `d` = the event ID (e.g., `<picture-event-id>`)
- `a` tag reference → `d` = the coordinate string (e.g., `30023:<author>:tokyo-guide`)
- `i` tag reference → `d` = the external content ID (e.g., `isbn:9784805311981`)

**Current code:** `publish_pin()` in `src/stores/social/pin_boards_store/publish.rs:56` builds tags for `A` (board refs), `e`/`a`/`i` (content ref), `title`, `image`, `t` — but **never adds a `d` tag**. Every pin nostr.blue publishes is spec-non-compliant.

`parse_pin_event()` in `mod.rs:500` also does not extract a `d` tag.

---

## 2. Kind 30067 boards: missing `d` tag drops valid events

`parse_pinboard_event()` in `src/stores/social/pin_boards_store/mod.rs:456-461`:

```rust
let d_tag = event
    .tags
    .iter()
    .find(|t| t.kind() == TagKind::SingleLetter(SingleLetterTag::lowercase(Alphabet::D)))
    .and_then(|t| t.content())
    .map(|s| s.to_string())?;  // ← ? returns None if no d tag
```

NIP-33 says a missing `d` tag is treated as `d=""` (empty string), not an invalid event. This code silently drops any kind 30067 event that lacks a `d` tag. A board with `d=""` is a valid parameterized replaceable coordinate `30067:pubkey:` — this parser rejects it.

**Fix:** Replace `?` with `.unwrap_or_default()`.

---

## 3. slugify produces empty d-tags for non-ASCII titles

`publish_pinboard()` in `publish.rs:10-12` generates the `d` tag via `slugify()`:

```rust
let d_tag = existing_d_tag
[...]

PR #354, opened 2026-08-15
Release 0.8.21: Deflock ALPR map, relay-boot race fix, thread scroll restore

patrickulrich:
### ✨ New: DeFlock ALPR camera map (`/deflock`)
Interactive Leaflet map of Automatic License Plate Reader cameras from
OpenStreetMap (Overpass API), modeled on maps.deflock.org.

- Full-screen map at `/deflock`; zoom-gated direction cones (DeFlock-accurate
  FOV-cone SVGs at z≥14, circles below; range/multi-value direction parsing)
- Leaflet.markercluster for tens of thousands of markers without jank
- IndexedDB persistence — revisiting an area next session is instant
- Bbox coverage tracking (skip refetch of covered viewports), parallel mirror
  racing across 3 Overpass endpoints, antimeridian-split queries
- Filter bar (zone/operator/text, debounced search) + OSRM directions panel
- Leaflet loader shared with Places via `utils/leaflet_shared.rs`

### ✨ New: PPQ Credit-ID import (AI chat)
Import an existing ppq.ai account by Credit ID; a dedicated API key is reused
or created automatically, and pasted keys are validated at import time.

### 🐛 Fix: relay lists seeded from disk at boot (#351)
Public relay lists (localStorage mirror + SDK DB) are seeded before the first
`connect()`, so `USER_RELAYS_APPLIED` flips synchronously — eliminating the
readiness race behind 44 `wait_for_user_relays` call sites. Includes
newest-wins seeding and a guard so a default-relay fallback can never
durably overwrite the user's real relay list.

### 🐛 Fix: threads open at top, Back restores position (#348)
`PENDING_SCROLL_TARGET` applied post-paint in NoteViewer with scroll-anchoring
suppressed during load; the pending-only guard prevents refresh re-fires from
yanking a mid-read user to the top.

### 🐛 Other fixes
- Pinboards NIP-01/33 compliance: derived `d` tags on pins, missing-`d`
  tolerance, Unicode slugify, media kind inference (#353)
- Profile metadata freshness: newest-wins kind-0 caching, outbox rescue,
  empty-name fallbacks, `lud06` zap fallback preserved
- Embedded comment/repost minicards are clickable
- PWA: build artifacts served network-first (no stale wasm), SW never
  registers in dev
- Map fixes from the #352 review round: popup CSS brace regression,
  per-map directions dispatch with quoted ids, antimeridian bbox coverage


src/components/places/map_container.rs
```diff
@@ -195,7 +174,7 @@ fn build_markers_js(id_json: &str, markers_json: &str) -> String {
                         ${{safeWeb ? '<div style="margin-bottom:3px;"><a href="' + safeWeb + '" target="_blank" rel="noopener" style="color:#a78bfa;text-decoration:none;font-size:12px;word-break:break-all;">🌐 ' + safeWeb.replace(new RegExp("^https?://"), "") + '</a></div>' : ''}}
                         <div style="display:flex;gap:6px;margin-top:8px;">
                             ${{s(m.naddr) ? '<a href="/' + s(m.naddr) + '" style="padding:5px 14px;border-radius:6px;background:transparent;color:#a78bfa;border:1px solid #7c3aed;cursor:pointer;font-size:12px;font-weight:500;text-decoration:none;">View Details</a>' : ''}}
-                            <button onclick="window.__placesRequestDirections(${{m.lat}},${{m.lng}},'${{name.replace(/'/g, "\\\\'")}}')"
+                            <button onclick="window.__requestDirectionsFor('${{mapId}}',${{m.lat}},${{m.lng}},'${{name.replace(/'/g, "\\\\'")}}','#7c3aed')"
```
goodmorningbitcoin: ## 🔴 High — Stored XSS via double-decoded `onclick` attribute

The `name` inserted here is HTML-escaped (`esc()` converts `'` → `&#39;`), and the `.replace(/'/g, "\\'")` guard is intended to JS-escape it. But the guard operates on the **already-escaped** string, where no raw `'` remains — it's dead code. The actual decode chain defeats both layers:

1. `esc()` turns `'` into `&#39;` and `"` into `&quot;`.
2. The popup HTML goes through `bindPopup` → `innerHTML`. The HTML tokenizer decodes character references **inside the attribute value** *before* handing it to the JS engine: `&#39;` → `'`, `&quot;` → `"`.
3. The JS parser then sees a raw `'` inside the `'${name}'` literal and the string terminates early.

So a place named `Li'l BBQ` produces `...'Li'l BBQ'...` → `SyntaxError` → dead Directions button. Worse, a crafted name like `'+alert(document.cookie)+'` decodes to a working JS expression → **arbitrary JS execution on click**. Place names here come from OSM (community-editable via Overpass) and user-published nostr places — i.e. attacker-controllable input.

The same pattern exists at line 193.

Notably, the **new deflock popup gets this right** — it interpolates only `c.osm_id` (a `u64`) and a JSON-quoted `mapId` into the onclick, which is why its test only asserts `mapId` quoting.

**Suggested fix:** stop interpolating names into inline JS entirely. Either

```js
// build with a data attribute + delegated listener:
<button class="dir-btn" data-lat="${m.lat}" data-lng="${m.lng}" data-name="${name}">
// ...
container.addEventListener('click', e => {
  const b = e.target.closest('.dir-btn');
  if (b) window.__requestDirectionsFor(mapId, +b.dataset.lat, +b.dataset.lng, b.dataset.name, '#7c3aed');
});
```

or JSON-quote at build time (`JSON.stringify(name)` yields a safely double-quoted JS literal that survives the attribute round-trip, since its internal `"` are entity-encoded by `esc`). While here, consider bumping the markercluster-style hardening to a shared helper so Places and Deflock can't drift again.

src/stores/profiles.rs
```diff
@@ -89,6 +104,21 @@ impl Profile {
         }
         truncate_pubkey(&self.pubkey)
     }
+    /// Whether the locally-known kind 0 is old enough (24h) to warrant a
+    /// background revalidation against indexers/outbox. Uses the source
+    /// event's `created_at` when known (Amethyst's freshness axis), else
+    /// falls back to when the Profile was fetched.
+    pub fn needs_revalidation(&self) -> bool {
```
goodmorningbitcoin, line 122: ## 🟡 Low — `needs_revalidation()` gates on event age, so stale profiles refetch on every view

`needs_revalidation` measures the **kind-0 event's** `created_at` (falling back to cache-insertion time). Most Nostr users' kind-0 events are older than 24h — profiles simply aren't republished daily. In `profile_viewer.rs` the stale path spawns `race_profile_metadata` on every mount, and a successful revalidation that finds *nothing newer* (the overwhelmingly common outcome for inactive users) records no "checked at" timestamp. Net effect: nearly every profile view of an infrequent poster triggers a full indexer+outbox race in the background, forever.

It's bounded (single-flight per view, strictly-newer guard on apply), so this is a chattiness/efficiency concern rather than a correctness bug — but it's new network load introduced by this PR.

**Suggestion:** add a `last_revalidated_at: Option<DateTime<Utc>>` to `Profile`, stamp it on both `Ok(Some(newer))` *and* `Ok(None)` revalidation outcomes, and gate `needs_revalidation()` on it:

```rust
pub fn needs_revalidation(&self) -> bool {
    let last_check = self.last_revalidated_at.unwrap_or(self.fetched_at);
    Utc::now().signed_duration_since(last_check).num_seconds() >= CACHE_TTL_SECONDS
}
```

That keeps the freshness axis (event `created_at` decides *whether* a newer snapshot can exist) but throttles *checking* to once per 24h per profile.

src/stores/social/pin_boards_store/publish.rs
```diff
@@ -53,12 +63,77 @@ pub async fn publish_pinboard(
...
     }
-    let mut tags: Vec<Tag> = vec![];
-    for board_addr in &input.board_addresses {
+    // Spec: the pin d tag is derived from the referenced content and is
+    // the first tag.
+    let d_tag = input.reference.d_tag_value();
+    // Re-pinning the same content replaces the prior pin event (addressable);
+    // merge its board refs so it stays pinned to earlier boards.
+    let board_addresses =
+        union_board_addresses(input.board_addresses.clone(), fetch_my_existing_pin_boards(&d_tag).await);
```
goodmorningbitcoin: ## 🟠 Medium — the board union makes it impossible to remove a pin from a board

`publish_pin` now merges the new selection with **every** board the existing pin references (`fetch_my_existing_pin_boards` → `union_board_addresses`). I searched the codebase: every `publish_pin` call site (`pin_new.rs`, `item_selector.rs`, `add_to_cookbook_modal.rs`) passes the user's *current* checkbox selection, and there is no un-pin/remove-from-board path for pins anywhere. Consequence: once content is pinned to board A, re-pinning it while deselecting A silently re-adds A via the union. The user cannot ever remove a pin from a board through the UI — the merge defeats their deselection.

The merge is clearly there for a good reason (don't lose earlier board refs when re-pinning from a different modal that only knows about its own selection) — but as written it also overrides an explicit full-selection republish.

**Suggestion:** distinguish the two intents. Either
- pass an explicit merge flag from modal call sites that only know a partial selection (`merge_boards: true`), while `pin_new.rs`'s full-form submit uses the selection as-is (`merge_boards: false`); or
- add a `remove_pin_from_board(board_addr)` flow that publishes with the board subtracted.

At minimum, document in the `PinInput`/`publish_pin` doc-comment that board membership is append-only today so the next reader doesn't assume deselection works.

goodmorningbitcoin (commented):
## 🧭 CodeRabbit-Style Review

**PR:** Release 0.8.21 — Deflock ALPR map, relay-boot race fix, thread scroll restore
**Scope:** 124 files, +4,712 / −695 · reviewed against `.coderabbit.yml` (assertive profile)

---

### 📋 Walkthrough

This release PR bundles five loosely-coupled workstreams:

1. **DeFlock ALPR map (`/deflock`)** — new Overpass-powered Leaflet map with zoom-gated FOV cones, markercluster, IndexedDB persistence, bbox-coverage dedup, 3-endpoint mirror racing, antimeridian-split queries, filter bar, OSRM directions. The service layer (`src/services/deflock.rs`) is genuinely well-tested (antimeridian math, direction parsing, bbox normalization all covered).
2. **Relay-boot race fix (#351)** — `relay/persistence.rs` implements a two-tier (localStorage mirror + SDK DB) synchronous boot seed before the first `connect()`, with a newest-wins fold and a defaults-overwrite guard for the metadata blob. The collect/write split to avoid holding a `WritableRef` across an `.await` is correct Dioxus practice.
3. **Thread scroll restore (#348)** — `PENDING_SCROLL_TARGET` applied post-paint with generation guards and scroll-anchoring suppression; the pending-only guard correctly prevents refresh re-fires from yanking a mid-read user.
4. **PPQ Credit-ID import** — proper input validation (UUID shape), key reuse-or-create with collision-avoiding names, redacted error bodies, replace-mode confirmation.
5. **Pinboards NIP-01/33 (#353)** — derived `d` tags, missing-`d` tolerance, Unicode slugify, kind inference — with tests.

Plus the `AUTH_STATE` → thread-safe pubkey mirror for the admission policy (verified: every write site now routes through `write_auth_state`; no drift path remains) and the quorum-EOSE NIP-78 migration (`connected_read_relay_count` correctly counts Connected + READ only).

### 🧵 Boot-seed sequence

```mermaid
sequenceDiagram
    participant Boot as run_post_login_init
    participant Disk as mirror + SDK DB
    participant Pool as relay pool
    participant Net as NIP-65 fetch
    Boot->>Disk: collect_relay_lists_from_disk (all awaits)
    Boot->>Pool: apply_seeded_relays_to_pool
    Boot->>Boot: write_seeded_relay_lists_to_signals (sync burst)
    Boot->>Pool: connect() — first wave includes user relays
    Boot->>Boot: USER_RELAYS_APPLIED flips early (no race)
    Boot->>Net: init_user_relay_lists (network refresh)
    Net-->>Boot: Ok → persist mirror (guarded) / Err → skip persist
```

### ⚠️ Findings (7 — 1 high, 2 medium, 4 low)

| Severity | Location | Issue |
|---|---|---|
| 🔴 High | `src/components/places/map_container.rs:177` (also :193) | Stored XSS via double-decoded `onclick` attribute |
| 🟠 Medium | `src/stores/relay/persistence.rs:97` | Defaults-overwrite guard doesn't cover the four NIP-51 lists |
| 🟠 Medium | `src/stores/social/pin_boards_store/publish.rs:134` | Board union makes un-pinning from a board impossible |
| 🟡 Low | `src/stores/profiles.rs:111` | SWR lacks a "last checked" stamp → per-view background refetch |
| 🟡 Low | `src/stores/relay/persistence.rs:191` | `limit(1)` trusts DB ordering the same file documents as unguaranteed |
| 🟡 Low | `src/services/deflock.rs:203` | `out body;` never populates `osm_timestamp`/`osm_version` |
| 🟡 Low | `src/stores/deflock_store.rs:68` | Unbounded bbox/camera cache; `CachedBbox.fetched_at` written but never read |

Details in inline comments below.

### ✅ Notable good work

- The antimeridian decomposition + exhaustive tests in `deflock.rs` — the wrapped-bbox invariants are actually asserted.
- `newest_relay_list_events` / `newest_metadata_by_author` folds explicitly replicate the SDK's `Ord for Event` semantics (descending `created_at`, ascending id tie-break) instead of trusting backend ordering — with tests proving last-iteration-wins is rejected.
- The `mapId`-quoted directions dispatch (with regression test in both map containers) fixes a real cross-map bug.
- PPQ response bodies stay redacted in errors (`redacted_response_body_len`), and the import flow never logs keys.

### 🏷️ Suggested labels

`rust` `ui` `wasm` `nostr-protocol` `security` `dependencies`

### 🎯 Review effort

**High** — 4.7k additions across 5 distinct subsystems. Recommend splitting future release PRs by workstream (the deflock feature alone would be a comfortably reviewable ~1.6k-line PR).

### 📝 Nitpicks

- `deflock_cache_db.rs`: `DeflockCacheDb::new()` is re-opened per fetch inside the fire-and-forget persist task — an IDB open per Overpass response. Cheap, but a `once_cell` handle would be free.
- `filter_bar.rs`: the pending debounce task isn't cancelled on unmount; the late `FILTERS` write is harmless (global signal) but worth a `use_drop` cancel for tidiness.
- `deflock_store.rs:59` — `is_viewport_covered` is containment-only (documented tradeoff); mid-zoom pans over tiled coverage will refetch. Fine for now; a rectangle-subtraction pass would cut Overpass load if it ever shows up in rate-limit logs.

### 🐰 Poem

*A map of watching eyes unfurled,*
*each cone a compass, quietly compiled;*
*but quotes within quotes within quotes decoded—*
*one string slipped its attribute, and ran.*
*Seed the relays from the disk, flip the gate at dawn—*
*the race is over before it began.*

---
*Review performed in CodeRabbit style per `.coderabbit.yml` (assertive, poem: true, request_changes_workflow: false — hence COMMENT, not REQUEST_CHANGES).*

src/stores/relay/persistence.rs
```diff
@@ -0,0 +1,449 @@
...
+    let Ok(pubkey) = crate::stores::nostr_client::get_cached_pubkey() else {
+        return;
+    };
+    if let Some(metadata) = USER_RELAY_METADATA.peek().as_ref() {
+        let existing = load_metadata(&pubkey);
+        if should_persist_metadata(metadata, existing.as_ref()) {
+            let _ = storage::set(&metadata_key(&pubkey), metadata);
+        }
+    }
+    let search = SEARCH_RELAYS.peek().clone();
```
goodmorningbitcoin, line 97: ## 🟠 Medium — the defaults-overwrite guard only protects the metadata blob; the four NIP-51 lists are persisted unconditionally

`persist_public_relay_lists` guards only `USER_RELAY_METADATA` via `should_persist_metadata` (the `updated_at == 0` defaults sentinel). The search/blocked/outbox/favorites writes at lines 97–104 run **unconditionally**, persisting whatever the signals currently hold.

The failure path exists: `init_nip51_relay_lists` writes `default_search_relays()` into `SEARCH_RELAYS` on *both* `Ok(empty)` and `Err(e)` (nip65.rs), and `default_favorite_relays()` into `FAVORITE_RELAYS` on both paths too. Sequence:

1. User has a custom kind 10007 search list → mirrored to `nostr.blue/relays/{pk}/search` in a prior session.
2. This boot: `run_post_login_init` calls `init_nip51_relay_lists` (auth_store.rs:1114) which fails or finds nothing (relay flake, offline) → signals now hold **defaults** (the disk seed's custom list was overwritten in the signal).
3. Line 1117 calls `persist_public_relay_lists()` unconditionally → defaults durably overwrite the user's mirrored search/favorites lists.

Note the mirror seed itself would have protected these lists *within* the session (`mirror_search` non-empty wins over the signal default? no — `init_nip51_relay_lists` writes the signal *after* seeding, so the default write wins in-session), but the mirror is then clobbered, so **every subsequent boot** serves defaults too. This is the exact class of bug the metadata guard was built to prevent — it just doesn't extend to the sibling lists.

**Suggested fix:** apply the same sentinel discipline to the four lists. Simplest: skip persisting a list when it equals the known defaults and the mirror holds a non-default value, or add a `defaults_origin: bool` to the write sites (set by `init_nip51_relay_lists`' default branches) and check it in `persist_public_relay_lists` the way `should_persist_metadata` checks `updated_at == 0`.

src/stores/relay/persistence.rs
```diff
@@ -0,0 +1,449 @@
...
+            dm_relays,
+            updated_at,
+        })
+    }
+}
+
+/// Query a single NIP-51 relay-list kind from the SDK DB.
+/// All four kinds (10006/10007/10012/10013) use `["relay", "url"]` tags.
+async fn collect_list_from_db(client: &Client, pubkey: &PublicKey, kind: Kind) -> Vec<String> {
+    let filter = Filter::new().author(*pubkey).kind(kind).limit(1);
```
goodmorningbitcoin: ## 🟡 Low — `collect_list_from_db` trusts `limit(1)` ordering that this same file documents as unguaranteed

The module doc and `newest_relay_list_events` were explicitly written because `database().query()` ordering "is not a documented newest-first guarantee across backends" — and then `collect_list_from_db` queries with `.limit(1)` and takes `events.into_iter().next()` anyway. If the NDB/IndexedDB backend returns oldest-first for a kind, the seed picks the **stale** NIP-51 list.

`newest_relay_list_events` exists and handles exactly this; the fix is a one-liner: bump the limit (say 10, matching `collect_metadata_from_db`), fold with the same newest-wins comparison (or extract a small generic helper over `created_at`/id ordering for single-kind queries). Low severity because the mirror (tier 1) usually has the fresh value and wins the seed anyway — this only matters when the mirror is empty (first login on a new device, or after 50k-cap eviction).

src/services/deflock.rs
```diff
@@ -0,0 +1,425 @@
...
+        lon: elem.lon,
+        operator: tags.get("operator").or_else(|| tags.get("surveillance:operator")).cloned(),
+        brand: tags.get("brand").or_else(|| tags.get("surveillance:brand")).cloned(),
+        directions,
+        direction_cardinal,
+        surveillance_zone: tags.get("surveillance:zone").cloned(),
+        mount_type: tags.get("camera:mount").cloned(),
+        ref_id: tags.get("ref").cloned(),
+        start_date: tags.get("start_date").cloned(),
+        osm_timestamp: None,
```
goodmorningbitcoin: ## 🟡 Low — `out body;` never returns timestamp/version, so `osm_timestamp`/`osm_version` are dead fields

The Overpass query uses `out body;`, which returns geometry + tags but **not** `timestamp`/`version` metadata (that requires `out meta;`). `parse_element` therefore hardcodes `osm_timestamp: None, osm_version: None`, the popup never shows them, and — because `AlprCamera` is `Serialize` and persisted to IndexedDB — every cached camera carries two dead `Option` fields.

Either add the metadata (change to `out meta;` and map `element.timestamp`/`element.version` with `#[serde(default)]` on the struct) or drop the fields until they're needed. The struct is also `PartialEq` + persisted, so dead fields cost a (tiny) IDB footprint per camera across potentially tens of thousands of entries.

src/stores/deflock_store.rs
```diff
@@ -0,0 +1,173 @@
...
+pub fn is_viewport_covered(viewport: &BoundingBox) -> bool {
+    FETCHED_BBOXES.read().iter().any(|f| {
+        viewport.south >= f.south
+            && viewport.north <= f.north
+            && viewport.west >= f.west
+            && viewport.east <= f.east
+    })
+}
+
+pub fn record_bbox(bbox: BoundingBox) {
```
goodmorningbitcoin, line 76: ## 🟡 Low — unbounded `FETCHED_BBOXES`/`CAMERAS` growth + `CachedBbox.fetched_at` is written but never read

Two related points:

1. `record_bbox` pushes forever — nothing prunes or merges bboxes, and the whole vec is re-persisted to IndexedDB after every fetch (one `put` per bbox in `insert_bbox`, called from the fire-and-forget task). A long session of country-scale panning accumulates hundreds of overlapping boxes; `is_viewport_covered` then linearly scans all of them on every 500ms poll tick, and the signal write re-renders anything subscribed.
2. `CachedBbox::fetched_at` is stamped (`now_millis()`) but **no code ever reads it** — it's dead until you implement TTL-based cache expiry.

Given ALPR data is close to static, an unbounded cache is defensible — but then `fetched_at` is dead weight, and a cheap containment-based merge (drop any stored bbox fully contained by a newly recorded one) would keep both the linear scan and the IDB row count flat. If you keep `fetched_at`, wire it to an expiry check at cache-warm time (e.g. ignore coverage from boxes older than N days) so stale-camera edits on OSM eventually propagate. Marking low because the practical failure mode is gradual perf decay, not incorrect data.

goodmorningbitcoin (commented):
CodeRabbit-style review — batch 2 (remaining findings: relay persistence, Overpass, deflock store).

src/stores/profiles.rs
```diff
@@ -794,9 +976,59 @@ pub async fn fetch_profiles_batch_native(
...
+        if !rescue_candidates.is_empty() {
+            log::info!(
+                "Outbox rescue: fetching metadata for {} profiles missing from indexers",
+                rescue_candidates.len()
+            );
+            use futures::stream::{self, StreamExt};
+            let rescued: Vec<(PublicKey, Metadata, u64)> = stream::iter(rescue_candidates)
+                .map(|pk| async move {
+                    let hex = pk.to_hex();
+                    match nostr_client::fetch_metadata_targeted(&hex, OUTBOX_RESCUE_TIMEOUT).await {
```
goodmorningbitcoin, line 1006: ## ⚠️ High — outbox rescue relay churn races `stream_events_immediate`'s snapshot → `RelayNotFound` kills whole streams

**Repro (field report on this branch):** open /notifications, tap a notification, go Back → page shows `❌ Nostr client error: relay not found` and stays dead (no retry path).

The chain, verified against nostr-relay-pool 0.44.1:

1. This rescue block connects ephemeral relays via `connect_ephemeral_relays` (coverage.rs:389–394) with `RelayOptions::new()` — **default flags `READ|WRITE|PING`** (relay/flags.rs:24-28). They are therefore returned by `client.relays()`.
2. `/notifications` load calls `prefetch_notification_authors` → `fetch_profiles_batch_native` → this block. After each targeted fetch, `cleanup_ephemeral_relays` → `client.force_remove_relay(url)` removes them mid-session.
3. On remount (Back nav), `stream_events_immediate` (streaming.rs:242-247) snapshots `client.relays()` filtered to Connected — snapshot includes any still-live ephemeral relay — then calls `stream_events_from(connected_urls, …)`. `stream_events_targeted` re-acquires the relays lock (an await point between snapshot and use). If cleanup removed a snapshotted URL in that window, `internal_relay` returns `Error::RelayNotFound` (pool/mod.rs:243) and **the entire stream fails** — one removed relay poisons all of them, because `?` propagates out of the loop over targets.
4. `stream_notifications` maps the error to `NostrBlueError::NostrClient` → the page's only error state. The load effect doesn't re-fire on error (only on auth/client-init changes), so the page stays broken until a full reload.

All five `stream_events_immediate` callers share the exposure: `routes/notifications.rs:1603`, `routes/home/feed_loaders.rs:437`, `routes/photos/home.rs:373,411`, `routes/polls/home.rs:111`, `routes/video/live_discovery.rs:143`.

**Suggested fix (pick one, cheapest first):**

```rust
// streaming.rs — after stream_events_from errors with RelayNotFound,
// fall back to the gossip stream instead of failing the whole load:
let stream = match client.stream_events_from(connected_urls.clone(), filter.clone(), timeout).await {
    Ok(s) => s,
    Err(e) if e.to_string().contains("relay not found") => {
        log::warn!("Snapshot raced pool removal, falling back to gossip stream");
        client.stream_events(filter, timeout).await?
    }
    Err(e) => return Err(e.into()),
};
```

Alternatively (root-cause fix): add ephemeral relays with a flag that excludes them from `client.relays()` snapshots — e.g. `RelayOptions::new().flags(RelayServiceFlags::GOSSIP)` — since targeted fetches address them by URL directly and don't consult flags. That also stops them inheriting pool subscriptions (pool/mod.rs add_relay: "if relay has READ flag, inherit pool subscriptions").

And in `routes/notifications.rs`, the `Err` arm at the stream call site could offer a retry — right now it's a dead page on a transient race.

src/stores/deflock_store.rs
```diff
@@ -0,0 +1,231 @@
...
+pub fn is_viewport_covered(viewport: &BoundingBox) -> bool {
+    FETCHED_BBOXES.read().iter().any(|f| contains(f, viewport))
+}
+
+/// Record a fetched bbox with containment-merge: skip insertion when an
+/// existing bbox already covers it, and drop stored bboxes fully contained
+/// by the new one (their coverage adds nothing). Without this the vec grew
+/// without bound during long panning sessions — every 500ms poll tick then
+/// linearly scanned all boxes and re-persisted them to IndexedDB.
+pub fn record_bbox(bbox: BoundingBox) {
```
goodmorningbitcoin, line 76: ## 📝 Nitpick — containment-merge drops bboxes from memory but not from IndexedDB

`record_bbox` now correctly drops stored bboxes fully contained by a newly recorded one — but `deflock_cache_db` has no delete/remove API, so the dropped rows persist in IDB indefinitely. The cache-warm path (map_container.rs:338-345) replays every stored row back through `record_bbox`, so behavior stays correct and the in-memory set self-heals; this is purely storage hygiene. If the IDB row count matters over long sessions, add a `delete_bbox(&self, bbox)`/`clear_bboxes()` and call it for the absorbed entries — or simpler, have the persist task rewrite the bbox store from the merged in-memory set instead of appending.

goodmorningbitcoin (commented):
## 🔁 Second-Pass Review — `bd69ab1f` (fix commit for round-1 findings)

**Verdict: all 7 round-1 findings verified fixed.** One new **high** finding on the interaction between this PR's outbox rescue and a pre-existing streaming race, detailed below with inline comments.

---

### ✅ Verified fixes (round 1 → round 2)

| # | Round-1 finding | Fix verified |
|---|---|---|
| R1 | 🔴 Stored XSS in Places popup `onclick` (double-decode) | ✅ Inline JS eliminated entirely — delegated container-level listener + `data-lat/lng/name` attributes; `mapId` interpolated as pre-quoted JSON; both button sites fixed; regression tests assert no `onclick=` remains |
| R2 | 🟠 NIP-51 defaults-overwrite on failed fetch | ✅ `apply_defaults_if_unseeded` — network result always wins; defaults only when the signal is unseeded; verified the boot order (`write_seeded_relay_lists_to_signals` at auth_store.rs:959 runs *before* `init_nip51_relay_lists` at :1114); tests cover all four outcomes |
| R3 | 🟠 Pin board union made un-pinning impossible | ✅ `PinInput.merge_boards` — selector modals (`item_selector.rs`, `add_to_cookbook_modal.rs` ×3 sites) pass `true` (additive), full pin form (`pin_new.rs`) passes `false` (authoritative); `resolve_board_addresses` + tests |
| R4 | 🟡 Profile SWR refetch-per-view | ✅ `last_revalidated_at` stamped on both outcomes (incl. `Ok(None)` nothing-newer) in both `profile_viewer.rs` call sites; `needs_revalidation` now gates on last-check time |
| R5 | 🟡 `collect_list_from_db` trusted `limit(1)` ordering | ✅ `limit(10)` + newest-wins fold using SDK `Ord for Event` semantics (`current < event` keeps the newer) |
| R6 | 🟡 Dead `osm_timestamp`/`osm_version` fields | ✅ Dropped from struct + parser + tests |
| R7 | 🟡 Unbounded bbox cache + dead `fetched_at` | ✅ `record_bbox` containment-merge (skip contained, absorb containing) with runtime-guard tests; `CachedBbox.fetched_at` dropped |
| N1/N2 | Nitpicks: per-fetch IDB open / uncancelled debounce | ✅ `get_or_open()` OnceLock handle; `use_drop` cancels the debounce task |

### 🧵 Not just the diff — boot-order dependency verified

The R2 fix only works because the disk seed writes the four NIP-51 signals *before* the network refresh runs. Confirmed both call sites in `auth_store.rs` (seed→signals at :959, `init_nip51_relay_lists` at :1114) — the "defaults only if unseeded" guard can actually fire on real boots.

---

### ⚠️ New finding: outbox rescue × snapshot race → `RelayNotFound` kills notification/feed streams

**File:** `src/stores/profiles.rs:976–1034` (rescue block, new in this PR) · `src/stores/nostr_client/streaming.rs:242–286` (pre-existing)

The chain, verified against nostr-relay-pool 0.44.1 source:

1. `/notifications` loads → `prefetch_notification_authors` → `fetch_profiles_batch_native` → **outbox rescue** (new) connects up to 10 ephemeral relays (`coverage.rs:393`, default flags `READ|WRITE|PING`) and `force_remove_relay`s them in `cleanup_ephemeral_relays` after each fetch.
2. User navigates Back → `/notifications` remounts → `stream_events_immediate` snapshots `client.relays()` (streaming.rs:242 — returns READ|WRITE-flagged relays, **ephemeral ones included**) and filters to Connected.
3. The snapshot releases the pool lock; `stream_events_from` → `stream_events_targeted` re-acquires it (an await point). If a rescue cleanup removed any snapshotted URL in that window, `internal_relay` returns `Error::RelayNotFound` (pool/mod.rs:243) — **the whole stream errors**, not just the missing relay.
4. `stream_notifications` propagates → `error.set(Some(e))` → the page shows `Nostr client error: relay not found` with an empty list and no retry path (the effect only re-fires on auth/client-init signal changes, not on error).

This matches a field report on this PR's branch: check a notification → go Back → page dead with `relay not found`.

Same exposure in all five `stream_events_immediate` callers: notifications, home feed, photos, polls, live discovery.

**Why this PR gets the flag:** the rescue is a *new, periodic* source of mid-session `force_remove_relay` churn (every profile batch with indexer-missing pubkeys, up to 10 relays × connect/remove per batch). Pre-existing removers (nip65 listener, health checks) had the same race but far lower frequency.

**Suggested fixes (any one suffices):**
- `stream_events_immediate`: re-check membership under one lock, or drop-not-fail on `RelayNotFound` per-URL (stream the intersection), or fall back to `stream_events` when `stream_events_from` errors with RelayNotFound;
- or `connect_ephemeral_relays`: add relays with a non-default flag set (e.g. `RelayOptions::new().flags(gossip-only)`) so they never enter `client.relays()` snapshots — they're only addressed by URL anyway (pool/mod.rs:243 doesn't flag-check on targeted calls);
- or in the notifications error path: surface a retry affordance instead of a dead page.

Not marking REQUEST_CHANGES per `.coderabbit.yml` (`request_changes_workflow: false`), but I'd fix this before release — it's a page-killing error triggered by routine profile prefetch on every notifications visit that has a cold-cache author.

---

### 📝 Nitpick (new): dropped bboxes linger in IndexedDB

`record_bbox` containment-merge drops stored bboxes when a covering one arrives, but `deflock_cache_db` has no delete path — dropped rows persist in IDB while the in-memory (and re-warmed) set stays flat. The cache-warm loop replays them through `record_bbox`, so behavior is correct; the rows are just dead weight. A `delete_bbox`/`clear_bboxes` on merge would keep storage matching memory.

---

**Tests:** the fix commit's new tests (`defaults_if_unseeded`, `merge_flag`, `bbox_merge`/`record_bbox`, `test_directions*`) pass on this head; full-suite run in progress at review time (949 tests at the prior head +4 new claimed). CI: Check & Lint pass; Android/Native pending.

*Second-pass review in CodeRabbit style per `.coderabbit.yml`.*

goodmorningbitcoin:
> ✅ **UPDATE — all six findings verified fixed in [`ee148ff6`](https://github.com/patrickulrich/nostr.blue/pull/354/commits/ee148ff69908d16c68006bec17ed39cb23349150) "fix(review): address PR #354 round-2/3 findings"** — audit details at the bottom of this comment.

## 🔐 Security Review — Cashu wallet (`release/0.8.21`)

Full review of the `/cashu` implementation (59 files, ~27k LOC): seed handling, token receive, proof lifecycle, melt path, NIP-60 sync, and NIP-61 nutzaps. Line numbers below refer to the `release/0.8.21` branch (these files are not part of this PR's diff, so no inline anchors).

**No critical findings — no fund theft, no key compromise, no injection paths found.** Four medium findings, two low. The core money-handling is genuinely well built.

---

### ✅ Verified sound

- **Seed handling** — deterministic derivation (nsec → SHA-256 domain-separated; NIP-07 → fixed-challenge schnorr sig), `ZeroizedSeed` wrapper zeroizes on drop, cache cleared on logout, `clear_seed_cache` wired into `logout()`. No seed/secret persistence to disk.
- **Melt path** — in-flight request persisted to storage *before* the operation starts (`lightning.rs:412` refuses to proceed if persist fails), per-mint operation lock, `try_operation_or_recover` re-syncs proof states with the mint after any failure and moves proofs to the correct local state (Spent/Pending/Spendable) based on the mint's answer — not on assumption.
- **NIP-60 sync** — all fetches author-filtered (`events.rs:275,302`), so no cross-pubkey proof injection. Balance sums use `checked_add` with real error paths on overflow. Kind-5 + `del`-field deletions both honored.
- **Nutzap validation** — `event.verify()` on every 9321, P2PK lock checked against our key with strict format validation (66-hex, `02` prefix, exact match), accepted-mint allowlist enforced, duplicate-claim guard via `add_pending_nutzap` returning `false` for known event ids.
- **Proof injection guard** — `create_ephemeral_wallet` dedupes by secret against the existing DB before injecting NIP-60 proofs into CDK.

---

### ⚠️ Findings

**M1 — HIGH · Nutzap receiving silently dies on every reload; pending nutzaps are memory-only**
`nutzap_signals.rs:15-35`, `nutzap.rs:574,698,711,1121`

`MY_NUTZAP_INFO` and `NUTZAP_ENABLED` are only ever set inside `publish_nutzap_info` (nutzap.rs:173-174). Nothing restores them at boot from the user's kind:10019 (which is on relays). Consequences after any page reload or app restart:

- `start_nutzap_subscription` errors out ("Nutzap info not published — enable receiving first"), so the live 9321 subscription never starts;
- `process_nutzap_event` **rejects every incoming nutzap** at nutzap.rs:711 ("Nutzap info not configured — cannot validate mint") when relays deliver old events;
- the backfill path `fetch_pending_nutzaps` also returns `Ok(0)` immediately when `MY_NUTZAP_INFO` is None (nutzap.rs:1121-1124);
- `PENDING_NUTZAPS` is a plain in-memory `GlobalSignal` — cleared on logout (`clear_nutzap_state`), lost on refresh.

Net effect: a user who enabled nutzap receiving receives zaps for exactly one session. Senders still see their kind:10019 and keep sending; nothing is stolen (the proofs stay P2PK-locked to the user's key), but funds sit unclaimed indefinitely with zero UI signal that receiving is dead — the inbox just says "not enabled". Recovery requires the user to re-publish 10019 *and* manually refresh the inbox. Suggested fix: at boot (after `init_wallet`), fetch own latest kind:10019, rebuild `MY_NUTZAP_INFO`/`NUTZAP_ENABLED` from it, restart the subscription, and run `fetch_pending_nutzaps`. Consider persisting `PENDING_NUTZAPS` (or deriving it from unredeemed 9321s + the 7376 "redeemed" history on relay).

**M2 — MEDIUM · `is_transient_error` substring matching misclassifies real mint errors as permanent**
`nutzap.rs:548-563`

`PERMANENT_PATTERNS` includes `"unknown"`, `"invalid"`, and `"keyset"` as bare substrings. CDK/http error strings like `unknown route`, `invalid state`, `Connection unknown` (or any mint-side 5xx whose body leaks into the error string) match and classify as permanent. When that happens mid-auto-redeem, the nutzap is stamped `Failed` and — because of M3 — has no retry path. The patterns are also matched against lowercased text where some entries are already lowercase/no-space variants (`"tokenalreadyspent"`), suggesting they were accumulated from observed strings rather than derived from the CDK error taxonomy. Suggested fix: match on `cdk::Error`/`cdk_sdk::Error` variants where available (the `Err` from `wallet.receive` is a typed error before it's stringified) and keep substring matching only as a last-resort fallback with tighter phrases.

**M3 — MEDIUM · `Failed` nutzaps have no retry button**
`nutzap_inbox.rs:241-252`

The Redeem button renders only `if is_pending`. A `Failed` nutzap (including ones failed via M2's misclassification, or via a genuinely transient network error that hit the permanent branch) shows a red badge with the error as a tooltip — and nothing else. `redeem_nutzap` itself works on any entry in `PENDING_NUTZAPS` regardless of status, so the fix is a one-line condition change plus keeping `is_redeeming` keyed the same way. Until M1 is fixed, a page refresh makes the entry vanish entirely, which is what makes this medium rather than cosmetic.

**M4 — MEDIUM · Receive path auto-adds unknown mints to the wallet with none of `add_mint`'s validation**
`internal.rs:138-152` (`get_or_create_wallet`) → `cashu_cdk_bridge::add_mint` (bridge, :54)

When receiving a token from a mint not yet in the wallet, `receive_tokens_with_options` → `create_ephemeral_wallet` → `get_or_create_wallet` falls through to the **bridge's** `add_mint`, which calls `repo.add_wallet_with_config` directly: no HTTPS scheme enforcement, no NUT-04/05 capability check, no keyset-collision check — all of which `mint_mgmt::add_mint` (mint_mgmt.rs:505-529) enforces for the manual add flow. The preview modal does surface `is_new_mint` to the user, so this is consent-adjacent rather than silent, but the receive flow is the one place where a *counterparty chooses the mint URL*, and it's the path with the weakest checks. Practical impact today is limited (proofs still have to verify against the mint's own keys to be worth anything, and bogus balances would fail at first spend), but a malicious mint can plant itself in the wallet state and appear in balance UI. Suggested fix: route the auto-add through the same validation as the manual flow (or at minimum the HTTPS + NUT-04/05 checks), and add the mint to the NIP-60 wallet event only after those pass.

**L1 — LOW · localhost bypass is a substring match**
`mint_mgmt.rs:510`

`url.host_str().unwrap_or("").contains("localhost")` matches `localhost.attacker.com` (and `notlocalhost.example`), allowing a plain-HTTP mint in the manual add flow. User-initiated, so low, but `host == "localhost" || host == "127.0.0.1"` (or `#[cfg(debug_assertions)]`-gated) is the correct check.

**L2 — LOW · Proof secrets in debug/warn logs**
`proof_recovery.rs:347` logs the full proof secret ("Proof {} is reserved at mint"); `recovery.rs:262` logs a truncated prefix (better). Browser-console logs outlive the tab in some setups and get attached to bug reports. Prefer logging a short hash prefix everywhere, as recovery.rs already does.

---

### Notable non-issues (checked, not problems)

- Token parsing is fully delegated to CDK with no `unwrap()` on parse in our code; the base64 padding auto-correct only *appends* `=`.
- `normalize_mint_url` defaults bare hosts to `https://` — an `http://` URL only survives if the token explicitly carries it (and M4 is the gap there, not normalization).
- Dioxus escapes all rendered strings; nutzap comments/memos render as text nodes only.
- The NIP-07 seed challenge uses `custom_created_at` + schnorr sig bytes — the wallet seed never touches extension-prompted arbitrary messages and is domain-separated from Nostr identity use.

*Security review of `release/0.8.21` in CodeRabbit style per `.coderabbit.yml`. All findings verified first-hand against the checked-out branch.*


---

## ✅ Audit of `ee148ff6` — all six findings verified fixed

| # | Finding | Verification |
|---|---|---|
| M1 | Nutzap receiving dies on reload | ✅ `restore_nutzap_state()` rebuilds `MY_NUTZAP_INFO`/`NUTZAP_ENABLED` from kind:10019 (SDK DB first, then relays via extracted `parse_nutzap_info_event`), restarts the 9321 subscription, backfills pending — fired fire-and-forget at the end of `init_wallet`, idempotent. Exactly the suggested fix. |
| M2 | `is_transient_error` over-matching | ✅ Compound tokens only (`tokenalreadyspent`, `keysetnotfound`, `alreadyspent`…), whitespace-normalized matching, **unknown defaults to transient** — correct asymmetry: wrongly-retryable costs one round-trip, wrongly-permanent strands funds. |
| M3 | No retry for Failed nutzaps | ✅ Inbox Redeem button renders for `Pending \| Failed(_)`; `redeem_nutzap` already worked on any entry. |
| M4 | Receive auto-add bypasses validation | ✅ `receive_tokens_with_options` validates unknown mints through the same `validate_mint_for_wallet` (HTTPS + exact-loopback + NUT-04/05) as the manual flow. The recursion hazard of validating inside `get_or_create_wallet` is real and correctly documented at both sites. |
| L1 | `contains("localhost")` bypass | ✅ Exact host match: `localhost`/`127.0.0.1`/`::1` only. |
| L2 | Full proof secret in logs | ✅ 8-char prefix with ellipsis. |

*Note: `Native Checks` (full `cargo test`) was still pending on this head at audit time; the commit claims 962 tests pass — confirm on the check run.*

src/components/places/map_container.rs
goodmorningbitcoin: ✅ **Verified fixed in `bd69ab1f`** (second-pass review). The inline `onclick` is gone entirely — replaced by a delegated container-level listener reading `data-lat/lng/name` attributes, so the place name never enters a JS parsing context at all (the correct fix, stronger than re-escaping). Both button sites converted, `mapId` interpolated as pre-quoted JSON, and regression tests assert no `onclick=` remains in the generated marker JS.

src/stores/profiles.rs
goodmorningbitcoin: ✅ **Verified fixed in `bd69ab1f`** (second-pass review). `last_revalidated_at` added to `Profile`, stamped on both revalidation outcomes (including `Ok(None)` nothing-newer) at both `profile_viewer.rs` call sites, and `needs_revalidation()` now gates on last-check time. Infrequent posters are checked at most once per TTL instead of on every view.

src/stores/social/pin_boards_store/publish.rs
goodmorningbitcoin: ✅ **Verified fixed in `bd69ab1f`** (second-pass review). `PinInput.merge_boards` distinguishes additive selector modals (`true`, all 4 call sites) from the authoritative full pin form (`false`), so deselection actually removes a board again. `resolve_board_addresses` + tests lock the contract in.

src/stores/relay/persistence.rs
goodmorningbitcoin: ✅ **Verified fixed in `bd69ab1f`** (second-pass review). `apply_defaults_if_unseeded`: a non-empty network result always wins; defaults land only when the signal is unseeded. Boot-order dependency verified (disk seed writes signals at `auth_store.rs:959`, before `init_nip51_relay_lists` at `:1114`), and runtime-guard tests cover all four outcomes. An offline boot no longer durably replaces custom NIP-51 lists with defaults.

src/stores/relay/persistence.rs
goodmorningbitcoin: ✅ **Verified fixed in `bd69ab1f`** (second-pass review). Now `limit(10)` + newest-wins fold using SDK `Ord for Event` semantics (`current < event` keeps the newer) — consistent with the file's own documented no-ordering-guarantee stance, matching `collect_metadata_from_db`.

src/services/deflock.rs
goodmorningbitcoin: ✅ **Verified fixed in `bd69ab1f`** (second-pass review). Both dead fields dropped from `AlprCamera`, the parser, and tests — no more persisting never-populated `Option`s to IndexedDB per camera.

src/stores/deflock_store.rs
goodmorningbitcoin: ✅ **Verified fixed in `bd69ab1f`** (second-pass review). `record_bbox` now containment-merges (skips contained bboxes, absorbs covering ones) with runtime-guard tests, and dead `CachedBbox.fetched_at` was dropped. Follow-up nit from the second pass (bbox rows dropped from memory but not from IndexedDB — no delete path) is noted on the round-2 thread and remains open.

src/stores/profiles.rs
goodmorningbitcoin: 🔁 **Status at `b35059db` (release head): still open.** `stream_events_immediate` is unchanged since this comment — the snapshot at streaming.rs:242 still feeds `stream_events_from` with no RelayNotFound fallback, and ephemeral rescue relays are still added with default READ|WRITE flags. Worth noting the new profiles-batching commit (`b35059db`) routes photo/video author metadata through `fetch_profiles_batch_native`, which *increases* how often the outbox rescue (and its connect/remove churn) fires — see the updated finding in the latest review round.

src/stores/deflock_store.rs
goodmorningbitcoin: 🔁 **Status at `b35059db`: still open** (nit, unchanged). `record_bbox` containment-merge drops bboxes from memory but `deflock_cache_db` still has no delete path, so absorbed rows linger in IndexedDB. Behavior stays correct via the cache-warm replay; purely storage hygiene.

goodmorningbitcoin (commented):
## 📋 Full-PR Review — 24 commits, head `b35059db`

Deep-review of the two commits not covered by earlier rounds; all prior threads annotated with verified-fixed status (round 1: all 7 ✅ fixed in `bd69ab1f`; round 2: both still open).

### `f5806baa` feat(mostro): protocol parity — ✅ clean

- **kind-14 chat envelope** (new `stores/mostro/chat.rs`): live subscription is author-pinned on `pub(K_sign)` — not `#p`-only, so it's flood-resistant; unwrap enforces outer author, single `p` tag, clock skew, content size bound, and inner-signer allow-list. Negative tests for wrong author and non-party signer.
- **Legacy 1059 dual-read**: mostro-core's giftwrap unwrap does not gate the inner signer — this code adds the explicit `allowed_signers` check on top (`chat.rs:81-83`). Correct. All call-site allow-lists derive from daemon-signed trade state only (trade key + counterparty for trade chat; admin + peer for disputes).
- **Restore fallback hardening** (`restore.rs`): the relay-poisoning defense is real — the decrypted seal must parse as kind-13, be authored by the configured daemon pubkey, and pass full id + signature verification; a forged-seal negative test covers it. The legacy unencrypted-rumor fallback is safe because the seal is authenticated before the rumor is read.
- **Ratings**: `Payload::Peer.reputation` arrives only via daemon-signed action messages — not client-forgeable. The upstream d-tag keying bug is documented and the relay fetch kept as best-effort.
- **Countdown**: `expiration_seconds` → hold-window → 900s matches mostrod's scheduler; display-only, no release-gating logic rides on it.
- No production `unwrap()`/`expect()` introduced (all new ones are `#[cfg(test)]`).

### `b35059db` perf(profiles): batch author metadata — one updated finding

The memo+queue pattern at all four call sites is correct (cache-version-reactive, `AlreadyBorrowed`-safe, no per-render fetch loops), and the app-shell drain is self-limiting (bumps only on fresh inserts; exhaustion gated at enqueue, errored chunks stay retryable). But the commit now routes every photos/videos feed author missing from `PROFILE_CACHE` into the outbox rescue — see the inline comment: this materially widens the round-2 `RelayNotFound` race, which is still unfixed at this head.

### Carry-over status at `b35059db`

| Finding | Status |
|---|---|
| `stream_events_immediate` RelayNotFound stream-kill | 🔴 open — exposure **increased** by `b35059db` (inline below) |
| deflock bbox rows linger in IndexedDB | 🟡 open nit |
| Cashu wallet findings (nutzap boot-restore, `is_transient_error` over-matching, no Failed-retry, receive auto-add bypass, `contains("localhost")`, proof-secret logging) | ⚪ untouched — last commit touching `src/stores/cashu/` anywhere in the repo is Jun 23; none of the six are addressed in this PR |

src/components/photo_card.rs
```diff
@@ -324,48 +323,30 @@ pub fn PhotoCard(
...
+    // the closure synchronously, and the inner `queue_profile_request` ->
+    // `bump_cache_version` -> `with_mut` would panic with `AlreadyBorrowed`.
+    let author_version = *crate::stores::profiles::PROFILE_CACHE_VERSION.read();
+    let _ = use_memo(use_reactive(
+        (&author_version, &author_pubkey_for_fetch),
+        move |(_v, pk): (u64, String)| {
+            if let Some(p) = crate::stores::profiles::get_profile(&pk) {
+                author_metadata.set(Some(p));
+            } else {
+                crate::stores::profiles::queue_profile_request(pk);
```
goodmorningbitcoin, line 341: 🔁 **UPDATED FINDING (round 2 → round 3): outbox-rescue churn is now triggered by every photos/videos feed load**

This commit adds four `queue_profile_request` sites (PhotoCard, VideoCard, LandscapeVideoCard, VertsPlayer comment authors). Each feed render enqueues every author missing from `PROFILE_CACHE`; the app-shell drain batches them into `fetch_profiles_batch_native`, whose outbox rescue (`profiles.rs:976-1034`) ephemeral-connects up to `MAX_OUTBOX_RESCUE = 10` authors' write relays at concurrency 3 (6s timeout each) and `force_remove_relay`s them after the batch.

That multiplies the exposure of the round-2 finding (thread at `profiles.rs:1006`): `stream_events_immediate` — used by notifications, home feed, photos, polls, and live discovery — still snapshots `client.relays()` and kills the **entire stream** with `RelayNotFound` if rescue cleanup removes a snapshotted URL inside the pool's re-acquire window (`nostr-relay-pool` 0.44.1 `stream_events_targeted`). Before this commit the trigger was the notifications author-prefetch; now any photo/video feed scroll fires it.

The batching itself is the right design — this is not a reason to revert. But it changes the severity math on the unfixed race:

1. **Preferred:** add rescue relays with non-default `RelayServiceFlags` (`RelayOptions` exposes `flags`) so they never appear in `client.relays()` — removes the churn from every snapshot at once.
2. **Also worth doing:** make `stream_events_immediate` fall back to `client.stream_events` (gossip) when `stream_events_from` fails, instead of propagating the error to a dead-end UI state.
3. Optional belt-and-braces: a small settle delay between rescue cleanup and the next `stream_events_immediate` snapshot.

Happy to pair on (1) — it's a contained change in `fetch_metadata_targeted`.

goodmorningbitcoin (commented):
## 📋 Full-PR Review — 24 commits, head `b35059db`

Deep-review of the two commits not covered by earlier rounds; all prior threads annotated with verified-fixed status (round 1: all 7 ✅ fixed in `bd69ab1f`; round 2: both still open).

### `f5806baa` feat(mostro): protocol parity — ✅ clean

- **kind-14 chat envelope** (new `stores/mostro/chat.rs`): live subscription is author-pinned on `pub(K_sign)` — not `#p`-only, so it's flood-resistant; unwrap enforces outer author, single `p` tag, clock skew, content size bound, and inner-signer allow-list. Negative tests for wrong author and non-party signer.
- **Legacy 1059 dual-read**: mostro-core's giftwrap unwrap does not gate the inner signer — this code adds the explicit `allowed_signers` check on top (`chat.rs:81-83`). Correct. All call-site allow-lists derive from daemon-signed trade state only (trade key + counterparty for trade chat; admin + peer for disputes).
- **Restore fallback hardening** (`restore.rs`): the relay-poisoning defense is real — the decrypted seal must parse as kind-13, be authored by the configured daemon pubkey, and pass full id + signature verification; a forged-seal negative test covers it. The legacy unencrypted-rumor fallback is safe because the seal is authenticated before the rumor is read.
- **Ratings**: `Payload::Peer.reputation` arrives only via daemon-signed action messages — not client-forgeable. The upstream d-tag keying bug is documented and the relay fetch kept as best-effort.
- **Countdown**: `expiration_seconds` → hold-window → 900s matches mostrod's scheduler; display-only, no release-gating logic rides on it.
- No production `unwrap()`/`expect()` introduced (all new ones are `#[cfg(test)]`).

### `b35059db` perf(profiles): batch author metadata — one updated finding

The memo+queue pattern at all four call sites is correct (cache-version-reactive, `AlreadyBorrowed`-safe, no per-render fetch loops), and the app-shell drain is self-limiting (bumps only on fresh inserts; exhaustion gated at enqueue, errored chunks stay retryable). But the commit now routes every photos/videos feed author missing from `PROFILE_CACHE` into the outbox rescue — see the inline comment: this materially widens the round-2 `RelayNotFound` race, which is still unfixed at this head.

### Carry-over status at `b35059db`

| Finding | Status |
|---|---|
| `stream_events_immediate` RelayNotFound stream-kill | 🔴 open — exposure **increased** by `b35059db` (inline below) |
| deflock bbox rows linger in IndexedDB | 🟡 open nit |
| Cashu wallet findings (nutzap boot-restore, `is_transient_error` over-matching, no Failed-retry, receive auto-add bypass, `contains("localhost")`, proof-secret logging) | ⚪ untouched — last commit touching `src/stores/cashu/` anywhere in the repo is Jun 23; none of the six are addressed in this PR |

src/stores/profiles.rs
goodmorningbitcoin: ✅ **Verified fixed in `ee148ff6`** — root cause + belt-and-braces, both correct.

**Root fix:** ephemeral rescue relays now carry `RelayServiceFlags::GOSSIP` only (`coverage.rs:407`). Verified against nostr-relay-pool 0.44.1 source: `pool.relays()` filters to `READ|WRITE` (`pool/mod.rs:217` → `relays_with_flag(READ|WRITE, Any)`), so rescue relays can no longer appear in the snapshot that feeds `stream_events_from`; `can_read()` includes GOSSIP (`relay/flags.rs:196`), so targeted URL calls still reach them. Connect/force-remove lifecycle unchanged, as claimed.

**Belt-and-braces:** `stream_events_immediate` now retries once with a fresh snapshot on typed `RelayNotFound` (`is_relay_not_found` matches the actual `client::Error::RelayPool(RelayNotFound)` variant first, string fallback), then falls back to the gossip stream instead of erroring the page. Other errors still propagate.

The device verification note in the commit (zero `relay not found` under feed loads + profile batches) matches the mechanism. Closing this out — thanks for the thorough fix.

src/components/photo_card.rs
goodmorningbitcoin: ✅ **Verified fixed in `ee148ff6`** (via the GOSSIP-flag root fix on the round-2 thread). Rescue relays are now excluded from `client.relays()` snapshots entirely, so the widened trigger surface from this commit's batching no longer produces snapshot/remove races. The batching design itself was correct as-shipped.

src/stores/deflock_store.rs
goodmorningbitcoin: ✅ **Verified fixed in `ee148ff6`.** `deflock_cache_db` gains `clear_bboxes`, and the persist path in `map_container.rs` now compacts the IDB store to mirror the in-memory containment-merge — absorbed rows are cleared and the surviving set rewritten, so IDB no longer accumulates absorbed bboxes. No residual.

